PR #1, opened 2015-07-29
Asd multiget2 variant using sendfile

domsj:


ocaml/src/asd_server.ml
```diff
@@ -360,8 +368,41 @@ let execute_query : type req res.
...
+                  let b = match blob with
+                    | Value.Direct s -> Asd_protocol.Value.Direct s
+                    | Value.OnFs (fnr, size) ->
+                       laters := (fnr, size) :: !laters;
+                       Asd_protocol.Value.Later size in
+                  b, cs))
+           keys
+         |> Lwt.return)
+      >>= fun (d,res) ->
+      (* AsdStatistics.new_multi_get stats d; *)
```
domsj: This duration doesn't contain everything yet ... that's why I commented it while working on this branch (and then forgot it afterwards).
But actually the same can be said for the duration for multiget on https://github.com/openvstorage/alba/pull/1/files#diff-8af52f72b48bb67ec73232134ae73a9fR369 , because that one doesn't take into account putting the value(s) on the wire.

Actually for all these commands we measure the duration twice, once here, and another time at 
https://github.com/openvstorage/alba/pull/1/files#diff-8af52f72b48bb67ec73232134ae73a9fR876 .
We probably only want to keep this last one (it includes the entire duration), and have it update the statistics.

I'll update that tomorrow.

domsj:
A similar change is still possible for the RangeEntries and Apply commands and may or may not follow later on

toolslive:
we don't use splice yet?!

domsj:
yes, we only use sendfile. I'll remove the wrapping for splice.

PR #16, opened 2015-08-19
OVS-2630: cpp client: distinct types for input vs. output stream exce…

toolslive:
…ptions

Also pass the "what" argument via const references to the exception constructors
while at it.


domsj:
looks good to me. we should some day enable travis to build the c++ client though

PR #22, opened 2015-09-01
Update policies

domsj:
A question came up while implementing this: should the nsm know which are the allowed policies and enforce that?


ocaml/src/albamgr_protocol.ml
```diff
@@ -1011,6 +1055,7 @@ module Protocol = struct
                       Wrap_u SetDefaultPreset, 25l, "SetDefaultPreset";
                       Wrap_u DeletePreset, 26l, "DeletePreset";
                       Wrap_u AddOsdsToPreset, 27l, "AddOsdsToPreset";
+                      Wrap_u UpdatePreset, 49l, "UpdatePreset";
```
domsj: change to 50l due to conflict with 'ports' branch

PR #25, opened 2015-09-07
License change

toolslive:


domsj:
looks like we're not going forward with this change after all ... closing.

PR #36, opened 2015-09-25
Verify namespace

domsj:


ocaml/src/alba_client_download.ml
```diff
@@ -34,3 +38,132 @@ let get_object_manifest'
     namespace_id object_name
     lookup_on_nsm_host
     ~consistent_read ~should_cache
+
+
+(* consumers of this method are responsible for freeing
+ * the returned fragment bigstring
+ *)
+let download_fragment
```
domsj: In hindsight extracting this one from alba_base_client wasn't really needed for this pull request...
(but still it makes alba_base_client thinner, so it's a good thing...)

Besides extracting it I also introduced Prelude.Error here

ocaml/src/repair.ml
```diff
@@ -0,0 +1,216 @@
...
+     alba_client # upload_object'
+                 ~namespace_id
+                 ~object_name
+                 ~object_reader
+                 ~checksum_o
+                 ~allow_overwrite >>= fun _ ->
+     Lwt.return ())
+    (let open Nsm_model.Err in
+     function
+     | Nsm_exn (Overwrite_not_allowed, _) ->
```
domsj: Pushed this error handling down into rewrite_object so not all it's callers have to do this

ocaml/src/repair.ml
```diff
@@ -0,0 +1,216 @@
...
+
+  let version_id = manifest.Manifest.version_id + 1 in
+
+  Lwt_list.map_s
+    (fun (chunk_id, chunk_location) ->
+
+     alba_client # get_namespace_osds_info_cache ~namespace_id >>= fun osds_info_cache' ->
+
+     let _, ok_fragments, fragments_to_be_repaired =
+       List.fold_left
```
toolslive: feels like a `List.partition`

ocaml/src/verify.ml
```diff
@@ -0,0 +1,166 @@
...
+               ~chunk_id ~fragment_id
+           in
+           let key = Slice.wrap_string key_string in
+
+           Lwt.catch
+             (fun () ->
+              osd_access # with_osd
+                ~osd_id
+                (fun osd ->
+                 if verify_checksum
```
toolslive: This is constant for as far as the maps are concerned; it's probably better to build a function
verify: osd_id -> key (* list *) -> [< `ChecksumMismatch | `Missing | `Ok ]
up front, that you pass down.

ocaml/src/repair.ml
domsj: The 2 resulting lists are of a different type though, so would require an extra transformation step after the partition... I'll leave it as is

PR #41, opened 2015-10-05
simpler and less inaccurate compilation

toolslive:


COMPILING.md
```diff
@@ -1,97 +1,44 @@
...
 
->To install the packages
->> - opam install package-name
+- Install the dependencies.
+The exact, up-to-date list of what you need can be found in the dockerfiles we
+use for CI.
 
->During the installation, you may be asked to upgrade a package. It's always preferable to pin the package (don't upgrade) specified by developer.
->> - opam pin package-name version
+*    [ubuntu/debian](./docker/alba_debian/Dockerfile).
```
domsj: Indentation / opsommingsteken might not be exactly what you had in mind here? (See rendered version of the .md file to see what I mean)
Also the '.' doesn't look like it's in its place there...

PR #52, opened 2015-10-22
Multicast independence

domsj:


ocaml/src/asd_server.ml
```diff
@@ -907,7 +911,14 @@ let asd_protocol
     write_extra fd
   in
   let rec inner () =
-    Llio.input_string ic >>= fun req_s ->
+    (match cancel with
+     | None -> Llio.input_string ic
+     | Some cancel ->
+        Lwt.choose
```
toolslive: pick?

domsj: should be Lwt.pick indeed

ocaml/src/osd_access.ml
```diff
@@ -40,33 +40,63 @@ class osd_access
...
+            "refresh_osd_total_used"
+            (fun () ->
+             Pool.Osd.factory
+               osd_buffer_pool
+               (Hashtbl.find osds_info_cache osd_id |> fst).Osd.kind
+             >>= fun (osd, closer) ->
+             Lwt.finalize
+               (fun () ->
+                osd # get_disk_usage >>= fun (used, total) ->
+                Osd_state.add_used_total osd_state used total;
```
toolslive: everywhere else, it's `_disk_usage` and here it's _used_total... It would make sense to use the same naming scheme. Also, the argument could be a tuple.

domsj: Makes sense, I'll adapt the name

ocaml/src/tools/networking2.ml
```diff
@@ -145,8 +143,21 @@ let make_server hosts port protocol =
...
-    inner ()
+      inner listening_socket in
+    let domain = Unix.domain_of_sockaddr socket_address in
+    let listening_socket = Lwt_unix.socket domain Unix.SOCK_STREAM 0 in
+    Lwt.finalize
+      (fun () ->
+       Lwt_unix.setsockopt listening_socket Unix.SO_REUSEADDR true;
+       Lwt_unix.bind listening_socket socket_address;
+       Lwt_unix.listen listening_socket 1024;
+       Lwt.choose
```
toolslive: pick?

PR #67, opened 2015-11-17
Ssl

toolslive:


fabfile/alba.py
```diff
@@ -79,29 +84,122 @@ def arakoon_start_(arakoon_config_file, base_dir, arakoon_nodes):
...
+        '--tls=%s,%s,%s' % ('%s/cacert.pem' % root,
+                            '%s/%s/%s.pem' % (root, name, name),
+                            '%s/%s/%s.key' % (root, name, name))
+    ])
+@task
+def arakoon_who_master(arakoon_cfg_file = arakoon_config_file,
+                       tls = 'False',
+                       env = env):
+    if _use_tls(tls):
+        arakoon_cfg_file = TLS['arakoon_config_file']
```
domsj:
```
def arakoon_who_master(arakoon_cfg_file = None,
                                            ....
    if arakoon_cfg_file is None:
      arakoon_cfg_file = _use_tls(tls) ? TLS['arakoon_config_file'] : arakoon_config_file
```
(ok, I admit that's not valid python)
the point is that using tls doesn't mean the method should ignore the specified config file

fabfile/alba.py
```diff
@@ -25,6 +25,9 @@
 
 os.environ['ALBA_CONFIG'] = './cfg/test.ini'
 
+def _use_tls(tls):
+    return tls == 'True' or tls == 'true'
```
domsj: not really needed, but I would add `... or tls == True`

fabfile/alba.py
```diff
@@ -237,11 +361,23 @@ def arakoon_stop():
...
+
 @task
 def proxy_start(abm_cfg = arakoon_config_file,
                 proxy_id = '0',
-                n_proxies = 1, n_others = 1):
+                n_proxies = 1,
+                n_others = 1,
+                tls = 'False'):
     proxy_id = int(proxy_id) # how to enter ints from cli?
+    ssl = bool(tls)
```
domsj: unused?

fabfile/alba.py
```diff
@@ -356,34 +509,50 @@ def is_local(osd):
             return True
         return False
 
-    def inner():
+    def inner(abm_cfg):
+        if _use_tls(tls):
+            abm_cfg = TLS['arakoon_config_file'] #TODO:run_tests_recovery ?
```
domsj: This logic doesn't belong here ... overriding an argument given by the caller is a bit weird.
I suggest lifting this part to the top of claim_local_osds.

domsj: also you may want to introduce a method `get_default_arakoon_cfg(tls)` and use that everywhere

fabfile/alba.py
domsj: maybe rename to 'truthy'?

fabfile/alba.py
```diff
@@ -459,7 +655,7 @@ def smoke_test(sudo = False):
 
     my_fuser = "fuser"
 
-    if sudo:
+    if sudo <> 'False':
```
domsj: `if truthy(sudo):`

fabfile/dev.py
```diff
@@ -457,13 +546,127 @@ def run_test_big_object():
...
+
+        t1_test = time.time()
+        delta_test = t1_test - t0_test
+        test_name = "flavour_%i" % flavour
+        results.append((test_name, result, delta_test))
+
+    failures = filter(lambda x: not x[1], results)
+    t1 = time.time()
+    delta = t1 - t0
+    # TODO: is there a better way than manually generate this?
```
domsj: There is already dump_junit_xml() in alba.py. It currently dumps some dummy data and should provide a good starting point for what you're doing here.

ocaml/src/albamgr_protocol.ml
```diff
@@ -181,11 +183,12 @@ module Protocol = struct
   end
 
   module Osd = struct
-    include Nsm_model.OsdInfo
+    (*include Nsm_model.OsdInfo*)
```
domsj: remove comment?

ocaml/src/albamgr_protocol.ml
```diff
@@ -219,12 +222,30 @@ module Protocol = struct
...
+            let ips'  = Option.get_some_default ips ips' in
+            let port' = Option.get_some_default port port' in
+            let real_port, use_tls =
+              let open Tiny_json in
+              match other' with
+              | None -> port', tls
+              | Some other' ->
+                 try
+                   let r = Json.parse other' in
+                   let tlsPort = Json.as_int (Json.getf "tlsPort" r)  in
```
domsj: I was already wondering why I saw no UpdateOsds2 ... now I know how you did it :)

ocaml/src/asd_server.ml
```diff
@@ -295,6 +295,47 @@ let get_value_option kv key =
...
+                  else buffer_size
+                in
+                Lwt_bytes.read fd_in        buffer 0 step >>= fun bytes_read ->
+                write_all socket buffer 0 bytes_read >>= fun () ->
+                loop (todo - bytes_read)
+              end
+          in
+          loop size
+       in
+       copy_using buffer
```
domsj: wrap this with `Lwt.finalize` to `Lwt_bytes.unsafe_destroy buffer`
(or maybe add `let Lwt_bytes.with_buffer length f = ...` and use that?)

domsj: hmm, or use Buffer_pool.with_buffer

ocaml/src/asd_server.ml
```diff
@@ -1225,29 +1281,33 @@ let run_server
...
+         ~mgmt
+         ~get_next_fnr
+         asd_id nfd ic)
+  in
+  let server_t = Networking2.make_server ?cancel hosts port ~ctx:None protocol in
+  let maybe_add_tls_server threads =
+      match tls with
+      | None -> threads
+      | Some tls ->
+         let tlsPort = Option.get_some_default (port + 500) configuredTlsPort in
```
domsj: I would prefer to have the +500 be done in asd_config.ml, and that this method receives an explicitly specified port. That way al the defaults can be found in that file.

fabfile/env.py
```diff
@@ -24,6 +24,9 @@ def env_or_default(key, default):
     else:
         return default
 
+def is_true(tls):
+    return tls == 'True' or tls == 'true'
```
domsj: `... or tls == True`

ocaml/src/tools/net_fd.ml
```diff
@@ -0,0 +1,44 @@
+type t =
```
domsj: missing license header

ocaml/src/remotes.ml
```diff
@@ -132,36 +122,45 @@ module Pool = struct
...
+
+         Asd_client.make_client buffer_pool ~conn_info (Some asd_id)
+         >>= fun (asd, closer) ->
+         let osd = new Asd_client.asd_osd asd_id asd in
+         Lwt.return (osd, closer)
+      | OsdInfo.Kinetic (conn_info', kinetic_id) ->
+         let conn_info = Asd_client.conn_info_from conn_info' in
+         Kinetic_client.make_client buffer_pool ~conn_info kinetic_id
+         >>= fun (kin, closer) ->
+         Lwt.return (kin, closer)
```
domsj: redundant lines

ocaml/src/albamgr_protocol.ml
```diff
@@ -1017,12 +1040,12 @@ module Protocol = struct
            Nsm_host.to_buffer
            Llio.int64_to)
     | ListAvailableOsds ->
-      Llio.counted_list_to Osd.to_buffer
+      Llio.counted_list_to (OsdInfo.to_buffer ~version:2)
```
domsj: Hmm, this is a bit unexpected. Due to the `~version:2` old clients will no longer be able to interpret the result of this call when talking to a new server.
Shouldn't that have shown up in the tests you've added?
The same remark applies to all places below where you added ~version:2 to the serialisation.
I expected ListAvailableOsds2, ListOsdsByOsdId2, ... operations to be added.

toolslive: There are no old clients for this call in the compatibility tests. only old servers.

ocaml/src/tools/net_fd.ml
toolslive: ack.

cfg/asd_demo.json
```diff
@@ -1,16 +1,23 @@
...
- // turn off syncing of batches;
- // don't uncomment this unless you know what you're doing. warranty void
- //, "__sync_dont_use" : false
+    "log_level":"debug",
+    // should be an absolute path
+    "home": "/tmp/asd_demo",
+    "node_id": "xxxxxxxxxx",
+    // optionally specify the ips to listen for incoming client connections
+    // "ips" : ["1.2.3.4", "2.3.4.4"],
+    "port": 10000,
```
domsj: document this is now optional, and that not setting it disable plain socket access

cfg/asd_demo.json
```diff
@@ -1,16 +1,23 @@
...
+    // optional asd_id is optional
+    "asd_id" : "a globally unique id for the asd"
+    // optional, we use 90% of the file system capacity
+    , "limit" : 90
+
+    // optional:
+    , "tls" : {"cert": "path/to/cert",
+               "key" : "path/to/key"
```
domsj: document port property

fabfile/alba.py
```diff
@@ -136,30 +254,43 @@ def create_loop_devices():
...
-          '__sync_dont_use' : False
-    }
-    if multicast:
-        cfg ['multicast'] = 10.0
-    else:
-        cfg ['multicast'] = None
+    if not restart:
+        limit = 90 if env['osds_on_separate_fs'] else 99
+        asd_id = "%i_%i_%s" % (port,node_id, local_nodeid_prefix)
+        cfg = { 'port' : port,
```
domsj: We shouldn't set this property when testing tls, otherwise we're not sure it will really be using tls

fabfile/alba.py
```diff
@@ -79,26 +83,128 @@ def arakoon_start_(arakoon_config_file, base_dir, arakoon_nodes):
...
+    cfg = _my_client_tls()
+    cmd.extend(['--tls=%s,%s,%s' % cfg])
+
+@task
+def arakoon_who_master(arakoon_cfg_file = arakoon_config_file,
+                       env = env):
+
+    tls = env['alba_tls']
+    if is_true(tls):
+        arakoon_cfg_file = TLS['arakoon_config_file']
```
domsj: I think we only want to set arakoon_cfg_file if the caller of this method did not supply a value, otherwise it's a bit sneaky.

fabfile/dev.py
```diff
@@ -81,20 +86,111 @@ def run_tests_ocaml(xml=False, kind = default_kind, dump = None, filter = None):
...
+    cmd_line = ' '.join(cmd)
     if dump:
-        cmd = cmd + " > %s" % dump
+        cmd_line += " > %s" % dump
 
-    where(cmd)
+    where(cmd_line)
+
+@task
+def run_tests_cli(xml = False):
```
domsj: nice addition, we needed this

toolslive: it's a start

fabfile/alba.py
toolslive: indeed.

fabfile/dev.py
```diff
@@ -457,13 +571,138 @@ def run_test_big_object():
...
+            with warn_only():
+                local ("which alba.0.6")
+                local ("pgrep -a alba")
+                local ("pgrep -a arakoon")
+                local ("which fuser")
+                local ("sudo fuser -n tcp 10000 8001")
+
+            raise
+
+    def deploy_and_test(old_proxy, old_plugins, old_asds):
```
domsj: these arguments (old_proxy, old_plugins, old_asds) are unused, so you're probably not testing what you intended to test...

(I went looking for a bug in run_tests_compat as I still think some operations are not backwards compatible, see my other comment https://github.com/openvstorage/alba/pull/67#discussion_r45275986 )

toolslive: I ended up not needing the parameters there, but the test asserts that the current deployment
passes a smoke test and a simple object CRUD.

domsj: Hmm, looks like only the old_proxy argument is unused, I was a bit too quick with my previous comment.
Yes, I can see the simple object CRUD, my point is that it's not tested with the old proxy.

toolslive: you're right.

ocaml/src/nsm_model.ml
```diff
@@ -108,20 +160,24 @@ module OsdInfo = struct
       buf
       errors
 
-  let from_buffer buf =
-    let ser_version = Llio.int8_from buf in
-    assert (ser_version = 1);
+  let to_buffer  buf t ~version =
+    match version with
+    | 2 -> _to_buffer_2 buf t
```
domsj: If we length prefix this then future expansions of osdinfo could be easier

toolslive: yes. will do so.

PR #70, opened 2015-11-20
propagate ip/port change into osd_info_cache

toolslive:


domsj:
Is the change to the test code still needed after the changes to osd_access.ml?
I think the test should be able to pass as it was, which is a stricter test than the new variant

ocaml/src/osd_access.ml
```diff
@@ -320,17 +320,43 @@ class osd_access
...
+                       Osd.Asd(ips,port, id), used, total
+
+                  in
+                  let osd_info' =
+                    Osd.{ osd_info with
+                          kind = kind';
+                          total = total'; used = used'
+                    }
+                  in
+                  let () = Osd_state.add_disk_usage osd_state (used', total') in
```
domsj: for kinetic devices this will result in stale info being propagated to the albamgr

toolslive: what stale info?

domsj: used, total.

There's a loop to actively refresh the used,total info by polling the osds (because we want to be multicast independent) at https://github.com/openvstorage/alba/blob/master/ocaml/src/osd_access.ml#L54 .
In case of kinetic this line right here would undo the info that was previously polled.
Actually, come to think of it: in that loop (L54) we should update osd_info.total & used, not here...

toolslive: https://github.com/openvstorage/alba/blob/master/ocaml/src/kinetic_client.ml#L329

domsj: Yea, I know there's still some TODOs related to kinetic.
Ok so looking more closely this code is ok when the bug in https://github.com/openvstorage/alba/blob/master/ocaml/src/osd_access.ml#L54 gets fixed (the osd_info should be updated there to reflect the new (used,total) info). Can you make that change in this PR?

toolslive: will do

PR #77, opened 2015-11-30
retrieve albamgr client config in more cases

toolslive:


domsj:
We should probably fix the (same) bug in https://github.com/openvstorage/alba/blob/master/ocaml/src/remotes.ml#L41 too? Or at least not pretend that we're being clever there?

Btw nice test suite addition

ocaml/src/proxy_server.ml
```diff
@@ -480,35 +484,71 @@ let proxy_protocol (alba_client : Alba_client.alba_client)
...
+type refresh_result =
+  | Retry
+  | Res of Albamgr_protocol.Protocol.Arakoon_config.t
+
 let refresh_albamgr_cfg
     ~loop
     albamgr_client_cfg
     (alba_client : Alba_client.alba_client)
     destination =
+  let retrieve_from_any_node_except () =
```
domsj: I think you meant to drop the `_except` suffix

toolslive:
currently there are quite a few `loose cannon` tests. I plan to aggregate them in a suite (xml results, jenkins,...) in the immediate future.

ocaml/src/albamgr_client.ml
```diff
@@ -509,10 +512,50 @@ let wrap_around (ara_c:Arakoon_client.client) =
...
+      | node_name :: rest ->
+         Lwt.catch
+           (fun  () ->
+            let node_cfg  =
+              Hashtbl.find node_hashtbl node_name
+            in
+            let open Albamgr_protocol.Protocol.Arakoon_config in
+            let cfg = from_node_client_cfg node_cfg in
+            Lwt_log.debug_f "retrieving from %s" node_name >>= fun () ->
+            Client_helper.with_client'
```
domsj: At some point in the past we considered it possible that a node would accept a connection and not do anything with it for a long time (due to some arakoon bug at the time?), which is why we added https://github.com/openvstorage/arakoon/commit/fe49c8b764130307a4384c5215d366fb8d5bc06b to arakoon. I suggest doing the same thing here to protect us from such a scenario.

toolslive: we were investigating a but and the addition was done under that hypothesis. IIRC.

PR #78, opened 2015-12-01
added show-namespaces cli call

toolslive:


domsj:
What are you trying to achieve? That the framework can get this info for all namespaces with just 1 cli invocation?
If that's the case then this PR isn't enough, the framework will still have to write some loop around the cli call.

toolslive:
~100 times less creation of alba clients (and connection pools to abm, nsm)

ocaml/src/alba.ml
```diff
@@ -246,6 +254,99 @@ let alba_show_namespace_cmd =
...
+             let first' = List.hd_exn acc_namespaces' |> fst
+             and finc' = false
+             in
+             inner (acc_count', acc_namespaces', acc_has_more')
+                   ~first:first' ~finc:finc' needed'
+         in
+         inner (0,[],true) ~first ~finc needed
+       in
+       list_all () >>= fun (count, namespaces) ->
+       Lwt_list.map_s
```
domsj: map_p ?

PR #88, opened 2015-12-03
Tests cleanup fds

domsj:


toolslive:
reset fd limit to 1024 ?

domsj:
ack

PR #94, opened 2015-12-07
additional benchmarks: upload_fragments & ranges

toolslive:


ocaml/src/osd_bench.ml
```diff
@@ -85,6 +86,92 @@ let sets (client:Osd.osd) progress n value_size period prefix =
...
+           ~object_id ~version_id
+           ~chunk_id ~fragment_id)
+        "" Checksum.NoChecksum true
+    in
+    let asserts = [ assert_namespace_active; ]
+    and updates = [ set_data;
+                   set_recovery_info;
+                   set_gc_tag; ]
+    in
+    client # apply_sequence Osd.High asserts updates >>= maybe_fail
```
domsj: Reuse Alba_client_upload.upload_packed_fragment_data instead of imitating it here?

PR #100, opened 2015-12-09
Partial read cache stats

toolslive:


ocaml/src/proxy_protocol.ml
```diff
@@ -228,12 +232,19 @@ module ProxyStatistics = struct
      in
      ns_stats.download <- _update ns_stats.download delta
 
-   let new_read_object_slices t ns ~total_length ~n_slices ~n_objects ~took =
+   let new_read_objects_slices
+         t ns
+         ~total_length ~n_slices ~n_objects ~mf_sources
+         ~fc_hits ~fc_misses
```
domsj: these are not used?

toolslive: oops

PR #102, opened 2015-12-09
run at most n things concurrently

toolslive:


ocaml/src/maintenance.ml
```diff
@@ -741,25 +753,29 @@ class client ?(retry_timeout = 60.)
...
-                    >>= fun (_, devices, _) ->
-                    List.iter
-                      (fun osd_id ->
-                       if not (Hashtbl.mem threads osd_id)
-                       then begin
-                           Hashtbl.add threads osd_id ();
-                           Lwt.async
-                             (fun () ->
-                              Lwt_extra2.ignore_errors
+                    Lwt_pool2.use
```
domsj: Push this throttling down to (or around) the `self # garbage_collect_device` call?

domsj:
`decommission_device` and some items in `handle_work_item` (specifically RepairBadFragment & IterNamespaceLeaf, not the others?) can use some throttling too...

domsj:
I would wrap all the calls to `Repair.rewrite_object` & `Verify.verify_and_maybe_repair_object` (to request a 'resource' first).
The question at https://github.com/openvstorage/alba/pull/102#discussion_r47234341 remains.

toolslive:
I don't think `garbage_collect_device` needs to be throttled more than it already does by itself.

domsj:
You're right that garbage_collect_device doesn't need any throttling, so I suggest removing it. For the rest: looks good to me.

PR #108, opened 2015-12-14
log fd on open/close; OSDs don't change kind

toolslive:


ocaml/src/remotes.ml
```diff
@@ -132,13 +132,13 @@ module Pool = struct
 
     type t = {
       pools : (Osd.id, osd_pool) Hashtbl.t;
-      get_osd_kind : Osd.id -> OsdInfo.kind Lwt.t;
+      get_osd_kind : Osd.id -> OsdInfo.kind Lwt.t; (* TODO: should not have _ Lwt.t signature  *)
```
domsj: disagree with the comment you added here: we may have to get that info from the albamgr, so the `_ Lwt.t` signature here is expected imo.

ocaml/src/remotes.ml
```diff
@@ -166,21 +166,28 @@ module Pool = struct
...
-              ~check:(fun _ exn ->
-                  (* TODO some exns shouldn't invalidate the connection *)
-                  false)
-              ~factory:(fun () -> t.get_osd_kind osd_id >>= factory t.tls_config t.buffer_pool)
-              ~cleanup:(fun (_, closer) -> closer ())
-          in
-          Hashtbl.add t.pools osd_id p;
-          p
+          begin
+            t.get_osd_kind osd_id >>= fun kind ->
```
domsj: `kind` contains the connection info, which could in fact change.
It looks like in osd_access we however do some effort to invalidate a pool if the connection info is deemed to be incorrect, so probably this optimisation is not breaking anything.

PR #110, opened 2015-12-14
travis

domsj:
implement workaround for https://github.com/travis-ci/travis-ci/issues/5285


domsj:
grmbl, now it seems to be failing for another reason. will look into that later.

toolslive:
original mirror works again...

PR #112, opened 2015-12-15
fix asd_crud test

toolslive:


ocaml/src/cli_asd.ml
```diff
@@ -185,7 +185,7 @@ let asd_multi_get hosts port tls_config asd_id (keys:string list) =
 
      client # multi_get ~prio:Osd.High (List.map Slice.wrap_string keys)
      >>= fun values ->
-     print_endline ([%show: (Bigstring_slice.t * Checksum.t) option list] values);
+     print_endline ([%show: (Lwt_bytes2.Lwt_bytes.t * Checksum.t) option list] values);
```
toolslive: was a type error.

domsj: Huh, that's weird ... I would've expected a type error??

toolslive: it used to be:
`Bigstring_slice.show : 'a -> string`

so we are  giving the wrong type, but the generated code handed over to type inference is polymorph.

PR #116, opened 2015-12-18
enable tcp keepalive on server sockets

domsj:


domsj:
updated the pull request, it now does a bit more than previously advertised.
- enable tcp keepalive with sane defaults (configurable)
- update to latest arakoon
- removed nocrypto dependency

PR #124, opened 2016-01-04
Test shrinking of abm + refresh abm config in maintenance process periodically

domsj:


toolslive:
I'll adapt in the ocaml setup

PR #126, opened 2016-01-19
Etcd

toolslive:
CI ok but shaky (rebalance_one, but that's a known issue)


ocaml/src/albamgr_test.ml
```diff
@@ -20,9 +20,9 @@ open Albamgr_protocol.Protocol
 
 let host = "127.0.0.1"
 let hosts = [ host ]
-let ccfg_ref = ref None
+let ccfg_url_ref = ref None
 
-let get_ccfg () = Option.get_some (!ccfg_ref)
+let get_ccfg_url () = Option.get_some (!ccfg_url_ref)
```
domsj: It probably would've been easier (required less changes) if this type had stayed `unit -> Albamgr_protocol.Protocol.Arakoon_config.t` iso changing to `unit -> Prelude.Url.t`?

ocaml/src/asd_config.ml
```diff
@@ -50,3 +49,19 @@ let read_cfg cfg_file =
...
+  Lwt_extra2.read_file cfg_file >>= fun txt ->
+  Lwt_log.debug_f "Found the following config: %s" txt >>= fun () ->
+  retrieve_cfg_from_string txt
+
+
+
+let retrieve_cfg cfg_url =
+  let open Prelude.Url in
+  Lwt_io.eprintlf "config %s" (show cfg_url) >>= fun () ->
+  match cfg_url with
```
domsj: I think we want a retrieve_from_url helper method iso doing this pattern match everywhere. (See also https://github.com/openvstorage/alba/pull/126/files#diff-15c6c580d31e8c48332322b747723af7R114 )

ocaml/src/proxy_server.ml
```diff
@@ -60,20 +58,30 @@ let albamgr_cfg_to_ini_string (cluster_id, nodes) =
...
+             s 0 (String.length s) >>= fun () ->
+           Lwt_unix.fsync fd) >>= fun () ->
+        Lwt_unix.rename tmp destination
+       )
+       (fun exn ->
+        Lwt_log.info_f ~exn
+                       "couldn't write config to destination:%s" destination
+       )
+  | (Url.Etcd etcd) as destination ->
+     Lwt_log.info_f "refresh NOT pushing to %s" (Url.show destination)
```
domsj: We need to think/discuss (e.g. with @khenderick ) what behaviour we want here...
It's possible that this code is executed in a proxy which is part of another ovs cluster

ocaml/src/tools/etcd.ml
```diff
@@ -0,0 +1,48 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*)
+open Lwt.Infix
+
+let parse_url url =
+  let host,port,path =
+    Scanf.sscanf url "%s@:%i/%s" (fun h i p -> h,i,p)
```
domsj: This means we only handle uri's with 1 etcd peer? Might be a problem down further down the road?

ocaml/src/tools/etcd.ml
```diff
@@ -0,0 +1,48 @@
...
+  let peers_s = String.concat "," (List.map (fun (h,p) -> Printf.sprintf "%s:%i" h p) peers) in
+  let cmd = ["etcdctl";
+             "--peers=" ^ peers_s;
+             "get";
+             path
+            ]
+  in
+  let cmd_s = String.concat " " cmd in
+  Lwt_io.eprintlf "cmd_s: %s" cmd_s >>= fun () ->
+  let command = Lwt_process.shell cmd_s in
```
domsj: fyi `Lwt_process.shell` basically undoes the earlier String.concat.
```
let command = ("", [| "etcdctl"; "--peers="^peers_s; "get"; path |]) in ...
```
should work just as fine

ocaml/src/proxy_server.ml
khenderick: The proxy has it's own key for storing the abm configuration, so I'd expect that the proxy keeps on updating its key like it does for when it's a file instead of an etcd key. It has it's own duplicated key because the code was - like @domsj stated - designed to be able to run a proxy on another cluster.

setup/prul.ml
```diff
@@ -0,0 +1,115 @@
+let get_some = function
```
domsj: missing license header

domsj: also prul isn't really the greatest name ...
and didn't you want to keep setup.ml as a single file for easy repl fiddling?

toolslive: I figured out to do both: repl fiddling & separate files. Feel free to suggest a better name for 'stuff that we use more than once, and doesn't really fit anywhere else'

travis.sh
```diff
@@ -108,6 +108,25 @@ install () {
     opam depext arakoon.1.8.12 orocksdb.0.2.1
     opam install ${OPAM_DEPENDS}
 
+    echo "Installing etcd based arakoon"
```
domsj: Arakoon 1.8.12 will already be installed by this point ... and then we'll do it again with the exact version we want, so we'll be wasting travis resources (and feedback time) with the current flow.
Remove arakoon from the OPAM_DEPENDS env var so this doesn't happen? (May require some more fiddling than that though, not sure...)

ocaml/src/albamgr_test.ml
toolslive: No. This was the thing that made the compiler show me the places that had to be adapted.

ocaml/src/cli_proxy.ml
```diff
@@ -42,55 +43,82 @@ module Config = struct
...
+       let url_from_cfg cfg =
+         let open Config in
+         let abm_cfg_url = Prelude.Option.map Url.make cfg.albamgr_cfg_url in
+         match cfg.albamgr_cfg_file, abm_cfg_url with
+         | None,Some u -> u
+         | Some f,None   ->
+            Lwt_log.ign_warning_f "albamgr_cfg_file is deprecated, please use albamgr_cfg_url";
+            (Url.File f)
+         | Some _,Some u ->
+            failwith "both albamgr_file and albamgr_cfg_url were confiured; just use later"
```
domsj: 'later' -> 'the latter' / 'albamgr_cfg_url'

ocaml/src/tools/etcd.ml
```diff
@@ -0,0 +1,48 @@
+(*
```
domsj: This file should be moved to arakoon and exposed in it's client lib?
And a similar remark for Prelude.Url ?

toolslive: That raises the question: do we want Arakoon to be the supplier of etcd functionality ?
I guess it's better to have an independent library for this.

ocaml/src/tools/etcd.ml
```diff
@@ -0,0 +1,48 @@
...
+  in
+  let cmd_s = String.concat " " cmd in
+  Lwt_io.eprintlf "cmd_s: %s" cmd_s >>= fun () ->
+  let command = Lwt_process.shell cmd_s in
+  Lwt_process.with_process_in
+    command
+    (fun p ->
+     let stream = Lwt_io.read_lines (p # stdout) in
+     Lwt_stream.to_list stream >>= fun lines ->
+     let txt = String.concat "\n" lines in
```
domsj: Hmm, first using read_lines and then concatenating again with \n ...

ocaml/src/albamgr_test.ml
domsj: Maybe I'm missing something, but if this didn't change then there would've been no need for the compiler to show you those places?

ocaml/src/tools/etcd.ml
domsj: A separate library would indeed be better, but (at least for now) I can live with it being in arakoon.
As long as we try to avoid code duplication it's fine for me

ocaml/src/tools/etcd.ml
domsj: @khenderick just confirmed on the chat that this is not a problem

ocaml/src/albamgr_test.ml
toolslive: yes, and we would end up trying to open files with the name `etcd://.........`

ocaml/src/asd_config.ml
toolslive: These are different configs, but I'll create a function
`retrieve_from_url: url -> (string -> 'a) -> 'a Lwt.t `
to split out the fetching.

ocaml/src/albamgr_test.ml
domsj: The url would've already been resolved to a Arakoon_config.t here, so that's no longer applicable for the code that uses this...
For what you're saying the change of type between alba_cfg_url and alba_cfg_file in cli_common already prevents that.

toolslive: The extra indirection is needed as fetching the config can only be done while inside the Lwt machinery.

domsj: Fwiw using Lwt_main.run that is not an insurmountable problem.
E.g. `let ccfg = Lwt_main.run (Alba_arakoon.config_from_url alba_cfg_url) in` should work just fine.

ocaml/src/tools/etcd.ml
toolslive: Can it be done better/shorter?

PR #127, opened 2016-01-21
Use fsyncs in asds instead of syncfs

domsj:


ocaml/src/asd_io_scheduler.ml
```diff
@@ -105,50 +119,152 @@ let run t ~fsync ~fs_fd =
...
+     then
+       begin
+         Lwt_buffer.harvest t.rocksdb_sync >>= fun wakers ->
+         t.sync_rocksdb ();
+         List.iter
+           (fun waker -> Lwt.wakeup_later waker ())
+           wakers;
+         Lwt.return_unit
+       end
+     else
```
toolslive: when can the `t.rocksdb_sync` buffer be empty ?

domsj: Anytime we get here it could be empty (due to the pick on wait_for_item above). I added the check so we wouldn't have to issue the sync to rocksdb unnecessarily

PR #130, opened 2016-02-03
Asd no blobs

domsj:
Add the dataloss guaranteed no_blobs option for asds (only to be used for testing purposes)

### DIFF
--- a/ocaml/src/asd_server.ml
+++ b/ocaml/src/asd_server.ml
@@ -82,16 +82,20 @@ module DirectoryInfo = struct
 
   type t = {
     files_path : string;
-    directory_cache : (string, directory_status) Hashtbl.t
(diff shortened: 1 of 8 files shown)


ocaml/src/asd_server.ml
```diff
@@ -82,16 +82,20 @@ module DirectoryInfo = struct
 
   type t = {
     files_path : string;
-    directory_cache : (string, directory_status) Hashtbl.t
+    directory_cache : (string, directory_status) Hashtbl.t;
+    no_blobs : bool;
   }
 
-  let make files_path =
+  let make ?(no_blobs = false) files_path =
```
toolslive: maybe spare some brain cycles and replace the double negative with ?(write_blobs = true)

domsj: ack, that is indeed clearer

PR #136, opened 2016-02-10
Redis logging

domsj:


docker/alba_centos7_jenkins/Dockerfile
```diff
@@ -96,7 +98,7 @@ RUN cd isa-l-2.14.0 && make install
 
 # install specific arakoon.
 RUN git clone https://github.com/openvstorage/arakoon.git
-RUN cd arakoon && git pull && git checkout tags/1.8.13
+RUN cd arakoon && git pull && git checkout 4d3a8913abfb5b70a2b8e8427c8e9019ddbe2776
```
toolslive: maybe wait a bit: I sense an imminent arakoon release.

ocaml/src/cli_common.ml
```diff
@@ -17,20 +17,80 @@ limitations under the License.
...
+
+  let hostname = Unix.gethostname () in
+  let component = Printf.sprintf "alba/%s" subcomponent in
+  let pid = Unix.getpid () in
+  let seqnum = ref 0 in
+  let logger = Lwt_log.make
+                 ~output:(fun section level lines ->
+                          let seqnum' = !seqnum in
+                          let () = incr seqnum in
+                          let ts = Core.Time.now () in
```
toolslive: core as a dependency in myocamlbuild.ml

docker/alba_centos7_jenkins/Dockerfile
domsj: ok, updated to arakoon 1.9.0

ocaml/src/cli_common.ml
domsj: ok, added

PR #154, opened 2016-02-23
Fadvise

toolslive:
- still have to figure out what to do with the headers & attribution.
- maybe we want fallocate on write too.


ocaml/src/asd_server.ml
```diff
@@ -217,24 +218,39 @@ module DirectoryInfo = struct
...
                 Lwt_extra2.write_all
                   fd
-                  s.buf s.offset s.length
-            ) >>= fun () ->
+                  s.buf s.offset len
+                >>= fun () ->
+                Lwt.return len
+            )
+            >>= fun length ->
+            Posix.lwt_posix_fadvise fd 0 length Posix.POSIX_FADV_DONTNEED
```
domsj: This fadvise should be done after syncing the blobs to disk. The sync is done in `post_write` 2 lines below.

ocaml/src/other/posix.ml
```diff
@@ -0,0 +1,19 @@
...
+                  | POSIX_FADV_NOREUSE
+                  | POSIX_FADV_WILLNEED
+                  | POSIX_FADV_DONTNEED
+
+external posix_fadvise: Unix.file_descr -> int -> int -> posix_fadv -> unit
+  = "_bs_posix_fadvise"
+
+let lwt_posix_fadvise fd pos len fadv =
+  let ufd = Lwt_unix.unix_file_descr fd in
+  Lwt_preemptive.detach
```
domsj: Hmm, the fadvise call isn't blocking? So no need for the overhead of pushing it to another thread?

ocaml/src/tools/object_reader.ml
```diff
@@ -43,9 +43,16 @@ let with_file_reader input_file f =
...
-     let object_reader = new file_reader fd stat.Lwt_unix.st_size in
-     f ~object_reader)
+      Lwt_unix.fstat fd >>= fun stat ->
+      let size = stat.Lwt_unix.st_size in
+      Posix.lwt_posix_fadvise fd 0 size Posix.POSIX_FADV_SEQUENTIAL
+      >>= fun () ->
+     let object_reader = new file_reader fd size in
+     f ~object_reader
+     >>= fun r ->
+     Posix.lwt_posix_fadvise fd 0 size Posix.POSIX_FADV_DONTNEED >>= fun () ->
```
domsj: For files with multiple chunks it would probably be beneficial to do this on a chunk level. In that case the fadvise would probably need to happen in the read method of the file_reader class.

domsj:
If the fallocate & fadv_sequential have impact we may want to do this too for the blobs in the fragment cache

toolslive:
I will add some configurability buttons

ocaml/src/fragment_cache.ml
```diff
@@ -247,7 +247,10 @@ class blob_cache root ~(max_size:int64) ~rocksdb_max_open_files =
   let db_path = root ^ "/db" in
 
   object(self :# cache)
-    val dirs = Asd_server.DirectoryInfo.make root
+    val dirs = Asd_server.DirectoryInfo.make
+                 root
+                 ~use_fadvise:false
+                 ~use_fallocate:false
```
domsj: `~fallocate:true` here? or make it configurable with true as the default

PR #160, opened 2016-03-03
as a different call

toolslive:


ocaml/src/cli_asd.ml
```diff
@@ -365,6 +365,100 @@ let osd_bench_cmd =
...
+            Lwt_io.printlf "%s : %s " long_id v
+          )
+      in Lwt_list.iter_s f results
+    in
+    let t () = 
+      with_alba_client
+        cfg_file tls_config
+        (fun alba_client ->
+          let open Nsm_model.OsdInfo in
+          alba_client # mgr_access # list_all_claimed_osds >>= fun (_n, osds) ->
```
domsj: This is inefficient when the call is used with only a fraction of the long-ids, and it doesn't work for non-claimed asds.
I think it would be better to add a multi-get-osds call and use that with all the specified long ids.

ocaml/src/cli_asd.ml
```diff
@@ -365,6 +365,100 @@ let osd_bench_cmd =
...
+      if to_json
+      then
+        begin
+          let result_to_json rs =
+            let x  =
+              List.map
+                (fun (long_id, statso) ->
+                  match statso with
+                  | None       -> long_id, `Null
+                  | Some stats -> long_id, Alba_json.AsdStatistics.to_yojson stats
```
domsj: It may make sense to wrap this result within a Alba_json.Result.t ?
That way we can later on - if desired - provide more info about which error happened while getting the statistics for a certain asd.

ocaml/src/cli_asd.ml
toolslive: Let's first see what this does.

PR #163, opened 2016-03-04
lower latency for small multiget2s

toolslive:


ocaml/src/tools/net_fd.ml
```diff
@@ -79,3 +79,13 @@ let read_all_lwt_bytes_exact target offset length = function
...
      >>= Lwt_extra2.expect_exact_length length
+
+
+let cork = function
+  | Plain fd -> Lwt_unix.setsockopt fd Lwt_unix.TCP_NODELAY false
+  | _ -> ()
+
+let uncork = function
+  | Plain fd -> Lwt_unix.setsockopt fd Lwt_unix.TCP_NODELAY true
+  | _ -> ()
```
domsj: I'm of the opinion that one should only match on `_` in exceptional circumstances...
In this case here for the Ssl case we may want to call flush ( https://github.com/savonet/ocaml-ssl/blob/master/src/ssl.mli#L378-L379 ), however I did not find that call in Lwt_ssl?
It may be useful to do the same perf tests with asds using ssl.

domsj:
Where else should we apply this change?
- arakoon server & client
- albamgr/nsmhost server & client (these are separate from arakoon...)
- asd client
- proxy server
- proxy clients (c++, ocaml)

toolslive:
you're exaggerating. most usage patterns are unaffected.

domsj:
Seems like I was indeed exaggerating. Your dismissive reply however doesn't help.
So let's scrap the asd client, proxy server and proxy ocaml client from that list.
The others look to me as if they are affected though.

PR #169, opened 2016-03-10
fix bug in auto-repair detection

domsj:
It was a bit too trigger happy causing unnecessary load on environments that are doing no reads

### DIFF
--- a/ocaml/src/automatic_repair.ml
+++ b/ocaml/src/automatic_repair.ml
@@ -34,23 +34,27 @@ let periodic_load_osds
         alba_client # osd_access # get_osd_info ~osd_id
         >>= fun (_, osd_state) ->
 
+        let write_test_blob () =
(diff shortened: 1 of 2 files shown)


ocaml/src/automatic_repair.ml
```diff
@@ -69,17 +73,26 @@ let periodic_load_osds
...
+                     (* seems like the test blob was not yet available on the osd
+                      * so let's put it there and retry... *)
+                     write_test_blob () >>= fun _ ->
+                     inner ()
+                  | _ ->
+                     Lwt.return ()
+                in
+                inner ())
                (fun exn ->
                 Lwt.return ())
```
toolslive: use `Lwt_extra2.ignore_errors`  ? (shows intent)

domsj: yes, updated

PR #170, opened 2016-03-10
Alba accelerated alba

domsj:
This pull request allows using e.g. a flash based alba as a fragment cache for a hard disk based alba.

Some more (related) changes are still to come in further pull requests:
- cache eviction
- extra statistics for the proxy
- configure cache on read/write behavior
- partial reads on the asds when no fragment cache is configured
- attempt to remove items from the fragment cache when removing/overwriting an object

### DIFF
--- a/ocaml/src/alba_client.ml
+++ b/ocaml/src/alba_client.ml
@@ -213,6 +213,33 @@ class alba_client (base_client : Alba_base_client.client)
     >>= fun _len ->
     Lwt.return !res
 
+  method download_object_to_bytes
+           ~namespace
+           ~object_name
+           ~consistent_read
+           ~should_cache
+         : Lwt_bytes.t option Lwt.t =
+    let res = ref None in
+    let write_object_data total_size =
+      let bs = Lwt_bytes.create (Int64.to_int total_size) in
+      res := Some bs;
+      let offset = ref 0 in
+      let write source pos len =
+        Lwt_bytes.blit source pos bs !offset len;
+        offset := !offset + len;
+        Lwt.return ()
+      in
+      Lwt.return write
+    in
+    self # download_object_generic
+         ~namespace
+         ~object_name
+         ~write_object_data
+         ~consistent_read
+         ~should_cache
+    >>= fun _len ->
+    Lwt.return !res
+
   method download_object_to_string'
            ~namespace_id
            ~object_name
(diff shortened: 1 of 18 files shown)


ocaml/src/alba_client_download.ml
```diff
@@ -172,6 +172,8 @@ let download_fragment
         >>= E.return)
      >>== fun (t_decompress, (maybe_decompressed : Lwt_bytes.t)) ->
 
+     fragment_cache # add namespace_id cache_key maybe_decompressed >>= fun () ->
```
domsj: this is a bug fix for the regression introduced by https://github.com/openvstorage/alba/pull/166

ocaml/src/alba_client.ml
```diff
@@ -213,6 +213,33 @@ class alba_client (base_client : Alba_base_client.client)
     >>= fun _len ->
     Lwt.return !res
 
+  method download_object_to_bytes
+           ~namespace
+           ~object_name
+           ~consistent_read
+           ~should_cache
+         : Lwt_bytes.t option Lwt.t =
+    let res = ref None in
```
toolslive: why do you need a reference here?

domsj: In this scope we don't yet know how big the Lwt_bytes.t needs to be, so we can't create it yet.
We need the reference here so we can return it later on (having it only available in write_object_data below is not enough).

It does look like I'm introducing a potential space/memory leak here (in case sth goes wrong during download_object_generic). I'll fix that.

ocaml/src/fragment_cache_alba.ml
```diff
@@ -0,0 +1,163 @@
...
+         with_namespace
+           bid
+           (fun namespace ->
+            client # nsm_host_access # with_namespace_id
+                   ~namespace
+                   (fun namespace_id ->
+                    base_client # download_object_slices'
+                                ~namespace_id
+                                ~object_name:(make_object_name ~bid ~name)
+                                ~object_slices:(List.map
```
toolslive: anonymous function 
as a parameter in an anonymous function 
as a parameter in an anonymous function
as a parameter in an anonymous function.

domsj: What exactly are you trying to say?
That in your opinion the code is unclear due to this?

toolslive: It boils down to this:
`(...)(...)` is heavier to process than `let f x = ... and x = ... in f x`
It's not that the code is unclear. It's just unduly heavy.

domsj: I think the 'unduly heavy' description is an exaggeration, regardless I've added another commit.

PR #171, opened 2016-03-16
Implement asd partial read

domsj:
and use it when no fragment cache is configured

### DIFF
--- a/ocaml/src/asd_server.ml
+++ b/ocaml/src/asd_server.ml
@@ -483,7 +483,7 @@ let execute_query : type req res.
         io_sched
         prio
         (fun () ->
-         Lwt_log.ign_debug_f "MultiGet for %s" ([%show: Slice.t list] keys);
+         Lwt_log.debug_f "MultiGet for %s" ([%show: Slice.t list] keys) >>= fun () ->
          (* first determine atomically which are the relevant Value.t's *)
          List.map
            (fun k -> get_value_option kv k)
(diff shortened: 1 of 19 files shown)


ocaml/src/asd_server.ml
```diff
@@ -581,6 +581,71 @@ let execute_query : type req res.
...
+                (fun acc (_, length) ->
+                 acc + 200 + length)
+                0
+                slices
+            in
+            perform_read
+              io_sched
+              prio
+              (fun () ->
+               Lwt_log.ign_debug_f "PartialGet for %s" (Slice.show key);
```
toolslive: why the `ign_` ?

ocaml/src/fragment_cache_test.ml
```diff
@@ -268,10 +281,55 @@ let test_long () =
...
   in
-  run_with_fragment_cache 20_000L _inner "test_long"
+  run_with_local_fragment_cache 20_000L _inner "test_long"
+
+let _test_lookup2 (cache : cache) =
+  let bid = 0l in
+  let key = "key" in
+  let size = 1_000_000 in
+  let value = Lwt_bytes.create size in
+  let value' = Lwt_bytes.to_string value in
```
toolslive: no equality on bigarray?

ocaml/src/asd_server.ml
domsj: I copy pasted that from the MultiGet(2) implementation, but it is indeed not needed, so removed the ign_.

ocaml/src/fragment_cache_test.ml
domsj: apparently there is, but for some reason I assumed there wasn't. adapting the code.

PR #172, opened 2016-03-16
cache on read/write configurability

domsj:


ocaml/src/fragment_cache.ml
```diff
@@ -22,20 +22,31 @@ open Lwt_bytes2
...
-    method close : unit -> unit Lwt.t
+    method virtual _add : int32 -> string -> Bigstring_slice.t -> unit Lwt.t
+    method add bid key blob (moment : [ `Read | `Write ]) =
+      let cache = function
+        | true -> self # _add bid key blob
+        | false -> Lwt.return_unit
+      in
+      match moment with
+      | `Read -> cache cache_on_read
+      | `Write -> cache cache_on_write
```
toolslive: isn't
`if (cache_on_read && moment = `Read) || (cache_on_write && momen = `Write) then ... ` simpler?
Come to think of it. It's a bit odd that the caller knows exactly which path (read or write) she's in and then encodes
it as a parameter for the callee to examine and determine the path from. Maybe it's better to have 2 methods, one for each path.

domsj: I think I wrote it like this to ensure exhaustive matching on 'moment'.

Regarding the other part of your comment ... there is indeed something a bit odd.
I think it's because the fragment cache shouldn't care where the add comes from. So pushed the decision wether or not to cache out to the caller in a new commit.

PR #174, opened 2016-03-22
move to arakoon 1.9.3

toolslive:


domsj:
Grep 1.9.2 and you'll see that you forgot to update a few places.
Actually there's still a few 1.9.1 and 1.9.0 to be found too...

toolslive:
you mean the alba_ubuntu_mini and alba_trisquel docker images ?
I can't really move those yet, as they install an alba together with a matching arakoon.

domsj:
Right, those can indeed not be moved yet.
I was thinking of travis.sh in the first place though, which still refers to arakoon 1.9.1.

PR #175, opened 2016-03-22
Osd bench for partial fetches

domsj:


ocaml/src/osd_bench.ml
```diff
@@ -54,6 +54,24 @@ let gets (client: Osd.osd) progress n value_size period prefix =
...
+
+let partial_reads (client : Osd.osd) progress n _value_size partial_fetch_size period prefix =
+  let gen = make_key period prefix in
+  let target = Lwt_bytes.create partial_fetch_size in
+  let do_one i =
+    let key = gen () in
+    client # partial_get
+           Osd.High
+           (Slice.wrap_string key)
+           [ 0, partial_fetch_size, target, 0 ] >>= function
```
toolslive: maybe we want to control the offset too?

domsj: I don't think the offset really matters for the performance, so that's why I left it out

PR #176, opened 2016-03-22
Alba cache eviction

domsj:
Completely untested for now ...
I think I'll need an env anyway to see if this is enough?
I could still make the 60s checkup period and 1% eviction configurable?


toolslive:
I would be surprised if it wouldn't need some (auto) tweaking. Since it's external, it should hurt when merged in.

PR #180, opened 2016-03-30
More asd performance related changes

domsj:


docker/alba_centos7_jenkins/Dockerfile
```diff
@@ -114,9 +114,13 @@ RUN yum -y install sudo
...
 RUN awk 'NR == 56 {next} {print}' /etc/sudoers >/tmp/__sudoers && mv /tmp/__sudoers /etc/sudoers
 
-RUN eval `${opam_env}` && opam update && \
-    opam depext orocksdb.0.2.2 && \
-    opam install -y orocksdb.0.2.2
+# install specific orocksdb
+RUN git clone https://github.com/domsj/orocksdb.git \
+    && eval `${opam_env}` \
+    && cd orocksdb \
+    && git checkout 8bc61d8a451a2724399247abf76643aa7b2a07e9 \
```
toolslive: We're losing the traceability of sub component versions. Maybe it's better to move to git submodules, combined with a more fine grained tracking, like what we do for the arakoon plugins.

PR #181, opened 2016-04-13
sendfile can EAGAIN

toolslive:
even when you wait for the socket to become available

### DIFF
--- a/ocaml/src/tools/lwt_extra2.ml
+++ b/ocaml/src/tools/lwt_extra2.ml
@@ -255,3 +255,24 @@ let read_file file =
        assert (read = len');
        Lwt.return ()) >>= fun () ->
   Lwt.return buf
+
+
+let copy_using fd_in fd_out size buffer =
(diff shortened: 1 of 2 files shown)


ocaml/src/tools/lwt_extra2.ml
```diff
@@ -255,3 +255,24 @@ let read_file file =
        assert (read = len');
        Lwt.return ()) >>= fun () ->
   Lwt.return buf
+
+
+let copy_using fd_in fd_out size buffer =
```
domsj: Hmm, didn't this loop already exist somewhere? (We probably want the code to be shared)

domsj:
looks good

ocaml/src/tools/lwt_extra2.ml
toolslive: I found a loop like this in the ssl sendfile part, but no where else...

domsj: yea, that should be the one

PR #182, opened 2016-04-13
Show namespaces speedup

domsj:


ocaml/src/tools/prelude.ml
```diff
@@ -525,6 +525,24 @@ let get_start_key i n =
       Some (serialize Llio.int32_be_to (Int32.of_int s))
     end
 
+module Result =
+  struct
+    include Result
```
toolslive: do we get anything in addition to a type definition?

domsj: no

PR #185, opened 2016-04-14
Rsocket

toolslive:


cpp/include/llio.h
```diff
@@ -81,6 +93,32 @@ class message_builder {
...
+                 [&](const char* buffer,
+                     const int len) -> void {
+                   os.write(buffer,len);
+                 });
+    os.flush();
+    if (!os) {
+      throw output_stream_exception("invalid outputstream");
+    }
+  }
+  /*
```
domsj: delete commented out code

cpp/src/examples/paranoia.cc
```diff
@@ -0,0 +1,56 @@
...
+  auto version = client -> get_proxy_version();
+    
+  int major = std::get<0>(version);
+  int minor = std::get<1>(version);
+  int patch = std::get<2>(version);
+  string commit = std::get<3>(version);
+  std::cout << "( " << major << ", " << minor << ", " << patch << ", "
+            << commit << ")" << std::endl;
+}
+int main(int /* argc */, char** /*argv */ ){
```
domsj: what value is this paranoia.cc file adding? what does it test? and if it tests something why not just make it a regular test?

cpp/src/lib/proxy_client.cc
```diff
@@ -269,5 +282,401 @@ Proxy_client::get_proxy_version() {
...
+    throw proxy_exception(errno, "connect");
+  }
+
+  _really_write((const char *)(&magic)  , sizeof(int32_t));
+  _really_write((const char *)(&version), sizeof(int32_t));
+
+}
+
+tuple<vector<string>, has_more>
+RDMAProxy_client :: list_namespaces(
```
domsj: All the functions below are almost identical to the implementation for the tcp proxy client.
I think this code duplication could've been avoided (as I already suggested in the chat when you started implementing this).
The duplication introduces opportunities for slight differences between the clients which we would rather avoid.
I think it would make sense to just have 1 proxy client class which only knows about `_reader` & `_writer` (and maybe a few other things), but it should know nothing about the transport method that is involved.

ocaml/src/alba_json.ml
```diff
@@ -58,7 +58,7 @@ module Osd = struct
       | OsdInfo.Asd (conn_info, asd_id)    -> "AsdV1", conn_info, asd_id
       | OsdInfo.Kinetic(conn_info, kin_id) -> "Kinetic3", conn_info, kin_id
     in
-    let ips, port, _ = conn_info in
+    let ips, port, _ , _ = conn_info in
```
domsj: maybe we want to expose this (useRdma) in the json?
(same could've been said for useTls earlier too though)

cpp/src/examples/paranoia.cc
toolslive: The file was here to explain a problem to @redlicha . This code worked, while almost identical test_client.cc SEGV'd. In the end it turned out to be a problem with the combination of (boost asio, c++ compiler and dynamic linking). So this file can be deleted.

ocaml/src/albamgr_client.ml
```diff
@@ -651,7 +651,12 @@ let make_client buffer_pool (ccfg:Arakoon_client_config.t) ~tcp_keepalive =
   find_master' ~tls ~tcp_keepalive ccfg >>= function
   | MasterLookupResult.Found (m , ncfg) ->
      let open Arakoon_client_config in
-     let conn_info = Networking2.make_conn_info ncfg.ips ncfg.port tls_config in
+     let transport = Net_fd.TCP in
```
domsj: Hmm, should this be considered a TODO?
Or is the idea here that if you want RDMA with arakoon then you'll have to use the preload lib anyway and it should all work fine?
Btw did you test the combination of preload + using rsockets towards the asds? (Just wondering if that might give issues)

ocaml/src/asd_server.ml
```diff
@@ -391,14 +392,44 @@ module Net_fd = struct
...
+                Lwt_bytes.read fd_in buffer 0 step >>= fun bytes_read ->
+                write_all socket     buffer 0 bytes_read >>= fun () ->
                 loop (todo - bytes_read)
               end
           in
           loop size
        in
        Buffer_pool.with_buffer Buffer_pool.default_buffer_pool copy_using
+    | Rsocket socket ->
+       let copy_using buffer =
```
domsj: You extracted a helper for this in https://github.com/openvstorage/alba/pull/181 which could be generalized a bit more to be usable here too

ocaml/src/asd_server.ml
```diff
@@ -1272,9 +1304,13 @@ let run_server
...
   Lwt_log.info_f "asd_server version:%s" Alba_version.git_revision     >>= fun () ->
-  Lwt_log.debug_f "tls:%s" ([%show: Asd_config.Config.tls option] tls) >>= fun () ->
-
-  let ctx = Tls.to_server_context tls in
+  Lwt_log.debug_f "tls:%s" ([%show: Asd_config.Config.tls option] tls_config) >>= fun () ->
+  Lwt_log.debug_f "transport:%s" ([%show: Net_fd.transport] transport) >>= fun () ->
+  let tls =
+    match tls_config with
+    | None -> (None: Tls.t option)
+    | Some tls_config -> failwith "todo"
```
domsj: what exactly is this todo about? you only mean to throw something when the (invalid) combination of rdma+tls is chosen?
also tls no longer works for the asds?
(probably I'm missing something...)

ocaml/src/discovery.ml
```diff
@@ -75,6 +75,7 @@ type record = {
     ips: string list;
     port :int option;
     tlsPort: int option;
+    useRdma: bool option;
```
domsj: this should be a plain `bool`, not an option

domsj:
you forgot to use the automated c++ code formatting using clang for you new code.
actually you made some whitespace only changes to existing code which conformed perfectly with the earlier agreed upon formatting.

ocaml/src/discovery.ml
toolslive: ok, but it makes our expression of what we see less accurate: we then have no way to express we did not see any `useRdma` field in the json message.

cpp/src/lib/proxy_client.cc
```diff
@@ -269,5 +275,471 @@ Proxy_client::get_proxy_version() {
...
+
+    read = rrecv(_socket, &buf[off], todo, flags);
+    if (read < 0) {
+      throw proxy_exception(read, "really_read.rrecv");
+    }
+    off += read;
+    todo -= read;
+    double t1 = _stamp_ms();
+    double delta = t1 - t0;
+    _request_time_left = _request_time_left - (int)delta;
```
domsj: _request_time_left update logic is not correct. When reading from the socket has to be done in multiple steps it will not wait as long as specified.

cpp/src/lib/proxy_client.cc
```diff
@@ -269,5 +275,471 @@ Proxy_client::get_proxy_version() {
...
+    }
+    sent = rsend(_socket, &buf[off], todo, flags);
+    if (sent < 0) {
+      throw proxy_exception(sent, "really_write.send");
+    }
+    off += sent;
+    todo -= sent;
+    double t1 = _stamp_ms();
+    double delta = t1 - t0;
+    _request_time_left = _request_time_left - (int)delta;
```
domsj: same remark as https://github.com/openvstorage/alba/pull/185/files#r62624982 regarding _request_time_left

cpp/src/lib/proxy_client.cc
```diff
@@ -269,5 +275,471 @@ Proxy_client::get_proxy_version() {
...
+    int rc = rpoll(&pollfd, nfds, _request_time_left);
+    ALBA_LOG(DEBUG, "rc=" << rc);
+    if (rc < 0) {
+      throw proxy_exception(rc, "really_write.rpoll");
+    }
+    if (rc == 0) {
+      throw proxy_exception(rc, "timeout");
+    }
+    sent = rsend(_socket, &buf[off], todo, flags);
+    if (sent < 0) {
```
domsj: We've seen that for sendfile (used with tcp sockets) it's possible that after waiting for a socket to become writeable it still gives back an EAGAIN/EWOULDBLOCK (can't remember exactly which one it was).
Should we be prepared to handle similar behaviour here too?

cpp/src/lib/proxy_client.cc
```diff
@@ -269,5 +275,471 @@ Proxy_client::get_proxy_version() {
...
+      throw proxy_exception(sent, "really_write.send");
```
domsj: We probably want to use errno here to build the error message

domsj: Hmm, the use errno remark seems applicable everywhere that functions from the rsocket library are called

cpp/src/lib/stuff.cc
```diff
@@ -18,102 +18,101 @@ limitations under the License.
...
-        a = t;
-    }
-    while (b != 0){
-        t = a % b;
-        a = b;
-        b = t;
-    }
-
-    return a;
+int gcd(int a, int b) {
```
domsj: 'gcd' and 'lcm' methods seem to be dead code
(was wondering why we have these and couldn't find any usages...)

docker/alba_centos7_jenkins/Dockerfile
```diff
@@ -126,6 +126,14 @@ RUN git clone https://github.com/domsj/orocksdb.git \
 # be uniform on shared lib locations
 RUN cp /usr/local/lib/librocksdb.so* /lib64/
 
+#for now, instll ordma manually
+RUN yum -y install librdmacm-devel
+RUN git clone https://github.com/toolslive/ordma.git \
+    && cd ordma \
+    && git checkout tags/0.0.2 \
```
domsj: update to current latest commit hash on ordma master as you might want this change too https://github.com/toolslive/ordma/commit/7ef1886d11fd7dd50fc1e307278e874006906ad7 ?
(same remark for other docker files & travis script)

domsj:
Still waiting for a reply on
https://github.com/openvstorage/alba/pull/185#discussion_r59712212
https://github.com/openvstorage/alba/pull/185#discussion_r59713318

Looks like you (silently) agreed with the other comments and made the relevant changes, thanks.

docker/alba_centos7_jenkins/Dockerfile
toolslive: That change is not really needed (the argument was wrongly ignored, and the field is set to `Closed` but it's only called with `Closed` as argument). So updating doesn't bring anything. meanwhile,
ordma is installable via opam. That would be an idea too.

cpp/src/lib/proxy_client.cc
toolslive: I considered this, but decided not to. Here's my rationale:

- I think it's wrong to try to generalize/simplify before the code is finished. In this case the timeouts still needed to be added and I was rather clueless on the effects that would bring. 
- not touching the tcp flavour of the client also means 0 chance for error on that client. 
- so we have 2 cases that look similar, with 3 cases I would certainly agree but now I'm not sure.
- lifting it in a superclass will probably make the volume of the code larger.

cpp/src/lib/proxy_client.cc
toolslive: indeed!

cpp/src/lib/proxy_client.cc
domsj: We should've had the rationale discussion earlier ... but here goes:
* Seeing the finished code I think it shows the suggested (or possibly other but similar) generalisation does make sense (e.g. the timeouts wouldn't be an issue)
* Using this kind of argument (not touching code means 0 chance for error) one could argue in favour of all kinds of standstill (against all changes). That's why I don't consider it a good argument, especially since 1) the changes would be rather straight forward and 2) that's what tests are for. Additionally not duplicating the code means there will be less chance for error on the new client, which is also important
* Agreed that having 2 cases would in the general case not be enough, here however I do think it's enough
* Contrary to your assertion I think the volume of code would actually get smaller (maybe I'm optimistic here, but honestly for me this isn't necessarily the most important argument)

cpp/src/lib/proxy_client.cc
toolslive: looking at the  source code:
https://github.com/weiny2/librdmacm-llnl/blob/master/src/rsocket.c
I'd be inclined to say no.

cpp/src/lib/stuff.cc
toolslive: moreover: it seems it was never alive.

cpp/src/lib/proxy_client.cc
domsj: Ok, so it's programming against an implementation iso the specification, but maybe that's ok here

PR #187, opened 2016-04-19
Asd disk quota

domsj:


cfg/asd_demo.json
```diff
@@ -16,7 +16,11 @@
     "port": 10000,
     // optional asd_id is optional
     "asd_id" : "a globally unique id for the asd"
-    // optional, by default we use 99% of the file system capacity
+
+    // optional, limit's the disk usage of the asd to the specified capacity
+    // by default the asd uses the entire file system capacity
+    , "capacity" : 1000000000000
```
toolslive: if configuration would not have been automatic, getting the number of zeros right would be hard.

ocaml/src/asd_server.ml
```diff
@@ -1277,13 +1305,12 @@ let run_server
...
   Lwt_log.debug_f "opening rocksdb in %S" db_path >>= fun () ->
-  let kv =
+  let db =
     Rocks_key_value_store.create'
       ~max_open_files:rocksdb_max_open_files
       ?recycle_log_file_num:rocksdb_recycle_log_file_num
       ~block_cache_size:(match rocksdb_block_cache_size with
-                         | None -> 0.0025 *. (Fsutil.disk_usage path
-                                              |> snd |> Int64.to_float)
+                         | None -> 0.0025 *. Int64.to_float !capacity
```
toolslive: explain the magic factor ? (you told me irl, but we tend to forget)

PR #190, opened 2016-05-03
fix for ovs-4456

domsj:
proxy startup takes too long when removing a large fragment cache

### DIFF
--- a/ocaml/src/fragment_cache.ml
+++ b/ocaml/src/fragment_cache.ml
@@ -79,24 +79,36 @@ let read_it fd ~len =
      Lwt_bytes.unsafe_destroy buffer;
      Lwt.fail exn)
 
-let rm_tree ~silent dir =
-      let cmd = Printf.sprintf "rm -rf %s" dir in
+let rm_tree dir =
+  Lwt_extra2.exists dir >>= function
+  | false -> Lwt.return ()
+  | true ->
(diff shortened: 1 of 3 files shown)


ocaml/src/fragment_cache.ml
```diff
@@ -79,24 +79,36 @@ let read_it fd ~len =
...
+  | true ->
+      let cmd = Printf.sprintf "mv %s %s-old-%f" dir dir (Unix.gettimeofday ()) in
       Lwt_log.debug_f "cmd:%S" cmd >>= fun () ->
       let command = Lwt_process.shell cmd in
       Lwt_process.exec command >>= fun status ->
       match status with
-      | Unix.WEXITED 0 -> Lwt_log.debug_f "cmd:%S ok" cmd
+      | Unix.WEXITED 0 ->
+         Lwt_log.debug_f "cmd:%S ok, now removing in the background" cmd >>= fun () ->
+         Lwt_process.exec (Printf.sprintf "mkdir %s" dir |> Lwt_process.shell) >>= fun _ ->
```
toolslive: don't we want to check this  process status?

domsj: My laziness didn't want to check it at first, but you're right, we should, and I will in another commit.

domsj: commit added

toolslive:
LGTM

PR #191, opened 2016-05-09
Allow getting asd disk usage with cli

domsj:


ocaml/src/cli_osd.ml
```diff
@@ -0,0 +1,168 @@
+(*
+Copyright 2015 iNuron NV
```
toolslive: 2016? (not that it matters)

toolslive:
lgtm

ocaml/src/cli_osd.ml
domsj: copy pasted that and forgot to update it...

PR #194, opened 2016-05-17
Travis updates

domsj:
- travis should fail if the unit tests fail (fixes a regression introduced in https://github.com/openvstorage/alba/pull/125)
- disable running the tests with tls (as they are broken since merging in #185)


toolslive:
ok. As a remark:the tls tests weren't running before the rsocket branch was merged in, due to a `true` `True` typo

PR #196, opened 2016-05-17
Agpl

toolslive:


domsj:
You missed the .mli files ( tools/alba_crc32.mli & tools/alba_wrappers.mli ).
The license headers in tools/fsutil.ml may be confusing? (But that was already the case before your pull request.)
For the rest it looks fine to me.

PR #198, opened 2016-05-18
removed superfluous file

toolslive:


domsj:
Travis build seems to have failed, and I can't really see why... maybe we should show more of the tail there? Or output xml from the unit tests and cat it after the test run? (will it still exit with an error then if not all tests passed?)
But considering the change in this pull request that failure is probably unrelated, so looks good to me.

PR #205, opened 2016-05-24
tail proxy log file when smoke test fails

domsj:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/132479420/log.txt

```
running test test_buffer_pool.buffer_pool_test.all
running test test_coordinator_master.maintenance_coordination.all
running test test_coordinator_participants.maintenance_coordination.all
running test test_fallocate.posix_test.all
running test test_example_config.fragment_cache_config_test.all
2016/05/24 07:34:29 203186 +0000 - testing-gce-2223555b-76c1-43a3-ab12-3821f608d0f6 - 14901/0 - alba/unit-tests - 917074 - debug - rselect.select: ([6;4],[],0.001398) begin
2016/05/24 07:34:29 203202 +0000 - testing-gce-2223555b-76c1-43a3-ab12-3821f608d0f6 - 14901/0 - alba/unit-tests - 917075 - debug - rselect.select => ([4],[]) end
[38;5;2m[shell] [0mpgrep -f '/home/travis/build/openvstorage/alba/ocaml/alba.native proxy-start --config /home/travis/build/openvstorage/alba/tmp/alba/proxies/00/proxy.cfg'
+expr 1
1
+false

travis_time:end:12bc3a29:start=1464075080770746122,finish=1464075269241468943,duration=188470722821
[0K
[31;1mThe command "./travis.sh script" exited with 1.[0m

Done. Your build exited with 1.
```

When the smoke test fails (because the proxy is no longer running) you currently have no idea why that is the case. Tailing the log file should help us get to the bottom of what's going on.

PR #206, opened 2016-05-24
cat testresults.xml on travis

domsj:
This will bring more clarity for what I actually wanted to know in https://github.com/openvstorage/alba/pull/205

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -179,8 +179,10 @@ script () {
             true
             ;;
         system2)
-            ${TEST_DRIVER} ocaml | tail -n256
-            expr $PIPESTATUS && false
+            ${TEST_DRIVER} ocaml | tail -n1000
+            X=$PIPESTATUS
+            cat testresults.xml
+            expr $X && false
             ;;
         disk_failures)
             ${TEST_DRIVER} disk_failures


travis.sh
```diff
@@ -179,8 +179,10 @@ script () {
             true
             ;;
         system2)
-            ${TEST_DRIVER} ocaml | tail -n256
-            expr $PIPESTATUS && false
+            ${TEST_DRIVER} ocaml | tail -n1000
+            X=$PIPESTATUS
```
toolslive: $PIPESTATUS[0] ?

domsj:
Ok, this change exposes the culprit for the sporadic test failures:
```

	<testcase classname="all.alba_test" name="test_create_namespaces" status="run" time="57.081463">
		<error message="Alba_client_errors.Error.Exn(6)"></error>
	</testcase>
```

Will have a deeper look at what's going wrong with this one at a later time...

toolslive:
ok.

travis.sh
domsj: It seems to work with just $PIPESTATUS, so I think in this case they are equivalent.

PR #214, opened 2016-05-31
allow running package integration tests with packages specified by urls

domsj:
and remove some unused cruft

(cleaning up the cruft in the jenkins job specification of alba_docker_deb/rpm_from_github is left as an exercise for another day)


toolslive:
ok.

PR #219, opened 2016-06-01
Alba as osd

domsj:
This pull request enables using a backend as an osd in another backend.
This functionality can be used to have e.g. (disk) repair/recovery on a datacenter level and on a global level.

2 todos currently remaining:
- alba_osd diskusage
- refreshing (in proxy/maintenance) & propagating (to global abm) of the alba_osd albamgr config

Nonetheless it seemed useful to already open the pull request so the review can be started.


ocaml/src/asd_client.ml
```diff
@@ -356,62 +356,67 @@ let make_client buffer_pool ~conn_info (lido:string option)  =
...
+        Lwt.return r
+
+      method get_exn prio (k:key) =
+        asd # multi_get ~prio [ k; ]
+        >>= function
+        | [ None ] -> Lwt.fail (Failure (Printf.sprintf
+                                       "Could not find key %s on asd %S"
+                                       (Slice.get_string_unsafe k) asd_id))
+        | [ Some (v, _) ] -> Lwt.return v
+        | _ -> assert false
```
toolslive: isn't it better to use the above `get_option` method ? 
(also, doesn't the asd object support a 'get' method?)

ocaml/src/asd_protocol.ml
```diff
@@ -62,17 +62,30 @@ module Blob = struct
...
+    | `S s1, `S s2 -> Slice.equal s1 s2
+    | `S s1, `B s2
+    | `B s2, `S s1 ->
+       Memcmp.equal'
+         s1.Slice.buf s1.Slice.offset s1.Slice.length
+         s2.Bigstring_slice.bs s2.Bigstring_slice.offset s2.Bigstring_slice.length
+    | `B s1, `B s2 ->
+       Memcmp.equal''
+         s1.Bigstring_slice.bs s1.Bigstring_slice.offset s1.Bigstring_slice.length
+         s2.Bigstring_slice.bs s2.Bigstring_slice.offset s2.Bigstring_slice.length
```
toolslive: maybe us a `let open Bigstring_slice in` ?

ocaml/src/asd_test.ml
```diff
@@ -71,7 +67,7 @@ let with_asd_client ?(is_restart=false) ?write_blobs test_name port f =
     begin
       capacity := 10_000_000L;
       Unix.system (Printf.sprintf "rm -rf %s" path) |> ignore;
-      Unix.mkdir path 0o777
+      Unix.system (Printf.sprintf "mkdir -p %s" path) |> ignore
```
toolslive: why?

ocaml/src/cli_asd.ml
```diff
@@ -582,11 +588,14 @@ let asd_statistics hosts port_o transport asd_id
                 | Some (claim_info, osd) ->
                   let conn_info =
                     let open Nsm_model.OsdInfo in
-                    let conn_info' = get_conn_info osd.kind in
+                    let conn_info' = match osd.kind with
+                      | Asd (x, _) -> x
+                      | Kinetic _ | Alba _ -> assert false
```
toolslive: maybe a function to get the Asd from the osd's kind ? (It's not the first time I see this inline extraction)

ocaml/src/alba_osd_test.ml
```diff
@@ -0,0 +1,90 @@
...
+     alba_client # mgr_access # get_alba_id >>= fun alba_id ->
+
+     let prefix = test_name in
+     alba_client # create_namespace ~namespace:prefix ~preset_name:None () >>= fun _ ->
+     let client = new Alba_osd.client alba_client ~alba_id ~prefix ~preset_name:None in
+     f (client # global_kvs))
+
+let test_set_get () =
+  test_with_kvs_client
+    "test_set_get"
```
toolslive: rename test to `test_set_get_delete` ?

ocaml/src/alba_arakoon.ml
```diff
@@ -19,13 +19,13 @@ but WITHOUT ANY WARRANTY of any kind.
...
+  type cluster_id = string [@@deriving show, yojson]
   (* TODO tls/ssl stuff *)
-  type node_name = string [@@deriving show]
+  type node_name = string [@@deriving show, yojson]
   type node_client_cfg = { ips : string list;
                            port : int; }
-                           [@@deriving show]
-  type t = cluster_id *  (node_name, node_client_cfg) Hashtbl.t
+                           [@@deriving show, yojson]
+  type t = cluster_id *  (node_name, node_client_cfg) Hashtbl.t [@@deriving yojson]
```
toolslive: didn't even know this was supported.

ocaml/src/alba_osd.ml
```diff
@@ -0,0 +1,281 @@
...
+                   ~object_reader:(let open Asd_protocol.Blob in
+                                   match value with
+                                   | Lwt_bytes s -> new Object_reader.bytes_reader s
+                                   | Bigslice s -> new Object_reader.bigstring_slice_reader s
+                                   | Bytes s -> new Object_reader.string_reader s
+                                   | Slice s -> new Object_reader.slice_reader s)
+                   ~checksum_o:(Some cs)
+                   ~object_id_hint:None
+                   ~fragment_cache:(alba_client # get_base_client # get_fragment_cache)
+                   ~cache_on_write:(alba_client # get_base_client # get_cache_on_read_write |> snd)
```
toolslive: If we're accessing this much (all the `alba_client # ... occurrences) from alba_client , 
only to hand it to alba_client_upload.upload_object'' there should probably be a method for this in alba_client.

ocaml/src/alba_osd.ml
```diff
@@ -0,0 +1,281 @@
...
+               Lwt.fail exn)
+    end
+  in
+  object(self :# Osd.osd)
+
+    method global_kvs =
+      get_kvs ~consistent_read:true prefix
+
+    method namespace_kvs namespace_id =
+      get_kvs ~consistent_read:false (to_namespace_name namespace_id)
```
toolslive: Isn't this the key for the namespace (calculated from the namespace_id) iso the name of the namespace ?

ocaml/src/alba_osd.ml
```diff
@@ -0,0 +1,281 @@
...
+                       ~preset_name
+                       () >>= fun _ ->
+           Lwt.return ()
+
+    method delete_namespace namespace_id _ =
+      alba_client # delete_namespace
+                  ~namespace:(to_namespace_name namespace_id) >>= fun () ->
+      Lwt.return None
+
+    method set_full _ = failwith "grmbl this method doesn't belong here."
```
toolslive: any idea what we want to do with this?

ocaml/src/alba_osd.ml
```diff
@@ -0,0 +1,281 @@
...
+    method get_version = Lwt.return Alba_version.summary
+    method get_long_id = alba_id
+    method get_disk_usage = Lwt.return
```
toolslive: the name should (in time) change. the implementation for alternatives can be done in a later pr. This one's large enough already.

ocaml/src/albamgr_protocol.ml
```diff
@@ -113,8 +113,12 @@ module Protocol = struct
     module Update = struct
       open Nsm_model.OsdInfo
       type t = {
+        (* asd/kinetic *)
```
toolslive: we might have been lifting out commonalities too early. Now with Alba osds, this is working against us.

ocaml/src/fragment_cache.ml
```diff
@@ -92,22 +92,32 @@ let rm_tree dir =
...
          begin
            let cmd = Printf.sprintf "mkdir %s" dir in
+           Lwt_log.debug_f "cmd:%S" cmd >>= fun () ->
            Lwt_process.exec (cmd |> Lwt_process.shell) >>= function
            | Unix.WEXITED 0 -> Lwt.return ()
-           | _ -> Lwt.fail_with (Printf.sprintf "rm_tree: cmd '%s' failed" cmd)
+           | Unix.WEXITED x ->
+              Lwt_log.info_f "rm_tree: cmd '%s' failed %i" cmd x >>= fun () ->
+              Lwt.fail_with (Printf.sprintf "rm_tree: cmd '%s' failed" cmd)
+           | _ ->
```
toolslive: Don't we want to know what happened ?

ocaml/src/fragment_cache.ml
```diff
@@ -92,22 +92,32 @@ let rm_tree dir =
...
+              Lwt_log.info_f "exotic" >>= fun () ->
+              Lwt.fail_with "exotic!"
          end >>= fun () ->
+         Lwt_log.debug_f "cucu1" >>= fun () ->
```
toolslive: debugging leftover?

ocaml/src/alba_osd.ml
domsj: I'm not quite following ... what do you mean with "key for the namespace"?
Are you referring to the prefix used for keys on the asds/kinetics?
I think the code here is correct, the intention is to calculate which namespace on the local backed should be used to store key/value pairs for a certain namespace (identified by it's id) from the global backend.

ocaml/src/alba_osd.ml
domsj: Yes, I think it should be removed from the osd class type and be asd specific instead. It requires changing a test, which I did not (yet) want to spend time on.
Implementing this method is not needed for regular alba functionality, this method is only used in the tests.

ocaml/src/alba_osd.ml
toolslive: I was referring to the name of the function `to_namespace_name` which confusing.

ocaml/src/asd_client.ml
domsj: Yes, this method, and the method above can indeed be made a bit shorter.

ocaml/src/asd_test.ml
domsj: This `-p` option for `mkdir` also creates any missing parent directories, which can be handy on some systems.
In particular this is convenient when I run the tests on my machine, for jenkins it's not needed.

ocaml/src/alba_osd.ml
domsj: How can `to_namespace_name` be a confusing name? The function takes as input a namespace id and gives back the name of the namespace to be used in the local backend.
It could be made more elaborate if that's what you're getting at...
Do you have any suggestions that don't confuse you?

ocaml/src/cli_asd.ml
domsj: This specific extraction happens exactly twice. I did it and it saved me exactly 0 lines of code.
I must say that it feels a bit like nitpicking if for the rest we pretend to not even care about the duplication damage inflicted upon the c++ proxy client in the rdma pull request.
(I may undo that damage someday though, it still bothers me.)

kvanhijf:
The ``--to-json``` for the ```add-osd``` call does not work as expected
Is it possible to also return the long-id of the newly added OSD in the returnvalue?

toolslive:
ok. merge it.

PR #225, opened 2016-06-03
GIOExecFile

toolslive:
unsolved:
- libgobjfs logs to stderr as we don't call  `google::InitGoogleLogging()` (calling it will make it log to file,
  which is also not what we want)
- libgogjfs will read it's config from a file, even if the ASD's config comes from etcd.


ocaml/src/asd_config_test.ml
```diff
@@ -0,0 +1,55 @@
...
+as published by the Free Software Foundation, in version 3 as it comes
+in the <LICENSE.txt> file of the Open vStorage OSE distribution.
+
+Open vStorage is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY of any kind.
+*)
+
+open OUnit
+open Asd_config
+let test_basic_config () =
```
domsj: Maybe I misunderstood, but I thought this test would put cfg/asd_demo.json under test?
Fwiw there's still an error in asd_demo.json (the 1_000_000 we talked about earlier, which was fixed on your branch, but then undone again when merging latest master?)

toolslive: No, this test was supposed to see how the jsonification of constructors with parameters would go.
Nonetheless, it's a good idea to put that file under test. Will do so.

domsj:
Still preparing some other comments (wanted to look at all code before actually making those comments), but I was already wondering the following:

What is conceptually the difference between the proposed changes and the current code base with as added change that the threads in the lwt worker pool would be pinned to cores?

domsj:
I'll answer my own question: one would be scheduling many IOs per thread concurrently (using aio), while the other would still be using plain old blocking reads on the worker threads...

toolslive:
with gIOExecFile we're bypassing the page cache entirely using O_DIRECT while in the "pure" version,
we're manipulating the page cache using fadvise.

ocaml/src/gblob_access.ml
```diff
@@ -0,0 +1,350 @@
...
+    let buf = Lwt_bytes.create 256 in
+    let rec loop () =
+      Lwt_unix.wait_read _reap_fd  >>= fun () ->
+      IOExecFile.reap _reap_fd buf >>= fun (n, ss) ->
+      (*Lwt_log.debug_f "reaped:%i" n >>= fun () -> *)
+      List.iter
+        (fun s ->
+          let completion_id = IOExecFile.get_completion_id s in
+          let ec = IOExecFile.get_error_code s in
+          let awake = Hashtbl.find _outstanding completion_id in
```
domsj: There's a race between passing a completion_id to IOExecFile and registering the callback in the _outstanding hashtable.
This could result in no callback being found on this line, which stops the inner harvest loop. That in turn results in consequent requests getting blocked. Maybe it explains the issue with hanging tests that you're still fighting?
I think `# next_completion_id` should already register the waker in the _outstanding hashtbl and return the sleeper along with the completion_id, this way you can't make any mistakes against it.
(_wait_for_completion then becomes obsolete)

ocaml/src/gblob_access.ml
```diff
@@ -0,0 +1,350 @@
...
+      (fun () ->
+        List.iter Fragment.free_bytes fragments;
+        Prelude.with_timing_lwt
+          (fun () ->IOExecFile.file_close handle)
+        >>= fun (took,()) ->
+        AsdStatistics.new_delta _statistics _FILE_CLOSE took;
+        Lwt.return_unit
+      )
+
+  method push_blob_data fnr len slices f =
```
domsj: `get_blob_data` is a better name than `push_blob_data`?

ocaml/src/blob_access.ml
```diff
@@ -0,0 +1,384 @@
...
+                    Lwt_unix.rmdir full_dir
+      end
+
+end
+
+let _FILE_OPEN     = 40l
+let _FILE_CLOSE    = 41l
+let _READ_BATCH    = 42l
+let _READ_WAIT     = 43l
+let _READ_BATCH_CB = 44l
```
domsj: Having these numbers in the same domain as a the asd commands is a bit of a hack ... I can live with it, but to prevent future accidents I think a note about this should be added near the Asd_protocol.ml command_map definition.

ocaml/src/blob_access.ml
```diff
@@ -0,0 +1,384 @@
...
+              then
+                Posix.posix_fadvise
+                  blob_ufd
+                  0 size
+                  Posix.POSIX_FADV_DONTNEED
+            in
+            Lwt.return_unit)
+          slices
+      )
+  method push_blob_data fnr size slices f =
```
domsj: push_blob_data and send_blob_data above seem to share a lot of code/structure, so I think it makes sense to extract out the commonalities.

ocaml/src/blob_access.ml
```diff
@@ -0,0 +1,384 @@
...
+    Lwt_extra2.with_fd
```
domsj: reuse `self # _with_blob_fd` here

ocaml/src/blob_access.ml
```diff
@@ -0,0 +1,384 @@
...
+
+
+  method get_blob fnr size =
+    Lwt_log.debug_f "getting blob %Li with size %i" fnr size >>= fun () ->
+    let bs = Bytes.create size in
+    self # _with_blob_fd
+      fnr
+      (fun fd ->
+         Lwt_extra2.read_all fd bs 0 size >>= fun got ->
+         assert (got = size);
```
domsj: There's a `Lwt_extra2.read_all_exact` method which combines the read and the assert

ocaml/src/blob_access_factory.ml
```diff
@@ -0,0 +1,59 @@
...
+  (r :> directory_access)
+
+let _service_handle = ref None
+
+let maybe_init_service gioexecfile =
+  match !_service_handle with
+  | None ->
+    begin
+      let service_handle = Gobjfs.IOExecFile.init gioexecfile in
+      let align_size = 4096 in
```
domsj: It's weird that this size is hard coded in ocaml, no?
I think this should come from somewhere in gobjfs itself... which in turn gets it from libaio or the kernel.

ocaml/src/blob_access_factory.ml
```diff
@@ -0,0 +1,59 @@
...
+  match engine with
+  | Config.Pure -> (new directory_info statistics config :> blob_dir_access)
+  | Config.GioExecFile conf_file ->
+     let service_handle = maybe_init_service conf_file in
+     (new Gblob_access.g_directory_info statistics config service_handle :> blob_dir_access)
+
+let endgame () =
+  match !_service_handle with
+  | None -> ()
+  | Some sh -> Gobjfs.IOExecFile.destroy sh
```
domsj: this should also reset the ref to None?

ocaml/src/gblob_access.ml
```diff
@@ -0,0 +1,350 @@
...
+
+  let _process_results get_ec function_name get_par results =
+    let bad = List.filter (fun t -> get_ec t <> 0l) results in
+    match bad with
+    | [] -> Lwt.return_unit
+    | t ::_  -> _throw_ex_from (get_ec t) function_name  (get_par t)
+  in
+
+  let event_channel = IOExecFile.open_event_channel service_handle in
+  let reap_fd = IOExecFile.get_reap_fd event_channel in
```
domsj: maybe event_fd is a better name for this?

ocaml/src/gblob_access.ml
```diff
@@ -0,0 +1,350 @@
...
+
+  initializer
+    begin
+      Lwt.ignore_result (self # _inner ())
+    end
+
+  val _next_id = ref 0L
+  val _outstanding =  (Hashtbl.create 16 : (int64, int32 Lwt.u) Hashtbl.t)
+  val _event_channel= event_channel
+  val _reap_fd = reap_fd
```
domsj: These 2 val's are unnecessary?
Make _next_id mutable iso a ref?

ocaml/src/gobjfs/gobjfs.mli
```diff
@@ -0,0 +1,84 @@
...
+  val init : string -> service_handle
+  val destroy: service_handle -> unit
+
+
+  type handle
+  val show_handle: handle -> string
+
+  type event_channel
+
+  val file_open: service_handle -> string -> Unix.open_flag list -> handle Lwt.t
```
domsj: This method doesn't need the 'a Lwt.t return type...
All it does is introduce more callbacks and (lwt) scheduling overhead... (and the potential for races such as the one I pointed out earlier)
Same remark for file_(write/read/delete/close) and close_event_channel...

ocaml/src/gobjfs/test.ml
```diff
@@ -0,0 +1,119 @@
+open Gobjfs
```
domsj: This file `ocaml/src/gobjfs/test.ml` looks like it's a duplicate of `ocaml/src/gobjfs/gobjfs_test.ml` (except for the missing license header), you probably want to remove it.

ocaml/src/gobjfs/gobjfs_test.ml
```diff
@@ -0,0 +1,121 @@
...
+in the <LICENSE.txt> file of the Open vStorage OSE distribution.
+
+Open vStorage is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY of any kind.
+*)
+
+open Gobjfs
+open Lwt.Infix
+
+let fill data=
```
domsj: `ocaml/src/gblob_access_test.ml` also has an almost identical copy of the `fill` and `check` functions that are present in this file.
They differ in the constant used for the modulo, one uses 256 and the other 100. I would recommend 251 because it's the biggest prime number smaller than 256.

ocaml/src/gblob_access.ml
```diff
@@ -0,0 +1,350 @@
...
+    let dir, _, file_path = self # _get_file_dir_name_path fnr in
+    self # ensure_dir_exists dir ~sync:sync_parent_dirs >>= fun () ->
+
+    let completion_id = self # next_completion_id  in
+    Lwt_log.debug_f "write_blob ~fnr:%Li completion_id:%Li" fnr completion_id >>= fun () ->
+    let len = Blob.length blob in
+    let alignment = 4096 in
+    let remainder = len mod alignment in
+    if remainder <> 0
+    then
```
domsj: What is the latest consensus regarding this path: will it remain necessary or not?

ocaml/src/gblob_access.ml
```diff
@@ -0,0 +1,350 @@
...
+      begin
+        Lwt_log.info_f
+          "ragged blob size: fnr:%Li len:%i (0x%0x) using ordinary write"
+          fnr len len
+        >>= fun () ->
+        Lwt_extra2.with_fd
+          file_path
+          ~flags:Lwt_unix.([ O_WRONLY; O_CREAT; O_EXCL; ])
```
domsj: As the file is not opened using O_DIRECT there should (at least optionally) an fadvise DONTNEED be done after the data is synced to disk.

ocaml/src/gblob_access.ml
```diff
@@ -0,0 +1,350 @@
...
+          ~perm:0o664
+          (fun fd ->
+            let extra = alignment - remainder in
+            let padded_size = len + extra  in
```
domsj: IIRC consensus was that this part (padding the blobs written to disk) would no longer be necessary.
And even if it was (or still is) necessary then making the copy would not be needed (just write the unchanged blob first, followed by some padding).

ocaml/src/gobjfs/gobjfs.ml
```diff
@@ -0,0 +1,356 @@
...
+      "gobjfs_batch_alloc"
+      (int @-> returning (ptr _batch))
+
+  let make fragments : t =
+    let len = List.length fragments in
+    if len = 0 then invalid_arg "empty batch";
+    let bp = _alloc len in
+    let b = !@ bp in
+    let a = getf b _array in
+    let rec loop ft = function
```
domsj: This looks like a List.fold_left

ocaml/src/blob_access.ml
toolslive: ok.

ocaml/src/blob_access_factory.ml
toolslive: I'll introduce a constant iso all the floating 4096's. But yes, it's weird we pull this out of thin air

ocaml/src/blob_access_factory.ml
toolslive: ok.

ocaml/src/gblob_access.ml
toolslive: I'll change it, but eventfd is an overloaded concept.

ocaml/src/gblob_access.ml
toolslive: ok.

ocaml/src/blob_access.ml
toolslive: ok. Eventually, protocol statistics should be separated from internal statistics.

ocaml/src/gblob_access.ml
toolslive: I considered get_blob_data first, but then the 'f' feels out of place. Anyway, I'll change.

ocaml/src/gblob_access.ml
toolslive: ok.

ocaml/src/gblob_access.ml
toolslive: the implementation on gIOExecFile is not there yet. you're right about the truncate.

ocaml/src/gobjfs/gobjfs.ml
toolslive: indeed.

ocaml/src/gblob_access.ml
toolslive: the path should go, and we should be totally ignorant about the alignment (we no longer need to know), but that code is not ready.

PR #234, opened 2016-06-10
Test link unlink alba osd scenario

domsj:
This exposed a bug in purging of osds and (a related) one in nsm update_manifest, both of which are now fixed


toolslive:
ok.

PR #238, opened 2016-06-13
Maintenance cache eviction

domsj:


ocaml/src/maintenance_config.ml
```diff
@@ -113,5 +163,15 @@ module Update = struct
...
-                             enable_rebalance'; }
+                             enable_rebalance';
+        cache_eviction_prefix_preset_pairs =
+          List.filter
+            (fun pair ->
+             not (List.mem
+                    pair
+                    remove_cache_eviction_prefix_preset_pairs))
+            (List.rev_append
+               add_cache_eviction_prefix_preset_pairs
```
toolslive: what if we add a pair that's already there?

domsj: Then objects in it will be evicted at twice the speed of others ... I'll make a change prevent this from happening

domsj: added a commit taking this comment into account

toolslive:
ok.

PR #251, opened 2016-06-20
Gitbook start

wimpers:


domsj:
Some files are added to the root of the repository, and others to the /docs subdir. Is that intentional?
I would expect all files you added to be part of /docs...

docs/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+Alba supports the following high-end set of features:
+* erasure coding
+* x-way replication
+* (optional) compression
+* (optional) encryption
+* partial reads
+* namespaces
+* [self healing](selfhealing.md) (repair, rebalancing)
+* ...
```
domsj: The current readme mentions a few more high-end features that you may want to add here?

domsj: This file also duplicates a lot of the current readme. You may want to remove some of the info from the current readme and just link to this one?

wimpers:
> Some files are added to the root of the repository, and others to the /docs subdir. Is that intentional? I would expect all files you added to be part of /docs...

Yes, docs is because on all other repo's it is like that. The GitBook and Summary are required for the GitBook content.

docs/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Namespace Manager
+The namespace manager is the remote API offered by the NSM host to manipulate most of the object metadata during normal operation. Its coordinates can be retrieved from the ALBA Manager by (proxy) clients and maintenance agents.
+
+
+## OSDs, ASDs and Seagate Kinetic drives
+ALBA supports 2 types of devices: Seagate Kinetic drives and ASDs. The Seagate Kinetic drives are ethernet-connected HDD with an open source object API designed specifically for hyperscale and scale-out environments. An ASD (ALBA Storage Device) is a our own pure software based implementation of the same concept on top of normal SATA disks: a Socket Server on top of RocksDb, offering persistence for meta data and small values, while large values reside on the file system. Every atomic multi-update is synced before the acknowledgement is returned. To have acceptable performance updates for multiple clients are written to the disk in batches. This is effectively a trade-off, increasing latency for throughput, which is perfectly acceptable for our use case (writes by the VM are already acknowledged when they hit the write buffer).
```
domsj: We now support 3 types of devices: an alba backend can be used as an osd too - resulting in the hierarchical alba with local & global recovery that we all know and love ;-).

docs/README.md
```diff
@@ -0,0 +1,62 @@
...
+Objects can also be encrypted or compressed. The combination of one or more policies, the type of compression and the encryption key is a **preset**.
+
+
+## ALBA Config files
+The ALBA configuration file is located at `/opt/OpenvStorage/config/alba.json`. For the maintenance process the node specific config file is located at `/opt/OpenvStorage/config/arakoon/<backend_name>-abm/<backend_name>.json`.
+For the rebalancer the node specific config file is located at `/opt/OpenvStorage/config/arakoon/<backend_name>-abm/<backend_name>-rebalancer.json`. More info on the ALBA config files can be found [here](../../Administration/Configs/alba.md)
```
domsj: - The link doesn't work?
- Configs are stored in etcd and not in the file system?
- There is no separate rebalancer (anymore)

docs/albaproxy.md
```diff
@@ -0,0 +1,114 @@
+# ALBA Proxy
+The ALBA proxy sits between the [Volume Driver](../VolumeDriver/README.md) and the [ALBA backend](README.md). It runs as a process on the Storage Router host and takes the SCOs coming from the Volume Driver and stores them according to a policy on the different OSDs of the backend.
+There is an ALBA Proxy per vPool which is available on the Storage Router. The port on which a vPool is listening for a certain vPool can be found in the config file of the proxy (`/opt/OpenvStorage/config/storagedriver/storagedriver/<vpool_name>_alba.json`).
```
domsj: The port on which a `proxy` is listening ...

docs/README.md
wimpers: > Configs are stored in etcd and not in the file system?
Any info where they are stored in etcd and how to interact with them?
Woudl should be that command?

domsj: `etcdctl get` & `etcdctl set` allow getting and setting values from etcd.
Somebody from the framework team should be able to tell you which keys are used for this... (e.g. @khenderick )
Fwiw: alba doesn't care which key is used here. It's a choice made by the framework. (Same can be said regarding service name, log file location, ... this is specific to how the framework uses alba - it's not something coming from alba itself)

khenderick: https://openvstorage.gitbooks.io/framework/content/docs/etcd.html

wimpers: Update the info so it points to the ETCD part on the Framework GitBook. We will tackle it there for now.

docs/README.md
```diff
@@ -0,0 +1,65 @@
...
+Objects can also be encrypted or compressed. The combination of one or more policies, the type of compression and the encryption key is a **preset**.
+
+## ALBA Config files
+The ALBA configuration files are stored in ETCD. 
+
+More info on the ALBA config files can be found [here](https://openvstorage.gitbooks.io/framework/content/docs/etcd.html)
+
+
+## ALBA Log files
+The log files for the ALBA maintenance process and the rebalancer can be found under `/var/log/upstart/ovs-alba-maintenance_<backend_name>.log` and `/var/log/upstart/ovs-alba-rebalancer_<backend_name>.log`.
```
domsj: There is no more separate rebalancer

docs/selfhealing.md
```diff
@@ -0,0 +1,87 @@
+# Self Healing
+ALBA has built-in self healing functionality. In case an ASD is broken, fragments on the ASD will be automatically reconstructed on remaining disks in the Backend.
+
+The self healing work consists out of several types of maintenance work:
```
domsj: Not all of the types of maintenance work listed below are related to self healing

PR #256, opened 2016-06-22
less annoying test failures on travis

domsj:
I changed to __retry_timeout setting of the maintenance process to speed up the alba_as_osd suite, but as a side effect it has been making CI (mostly travis but also jenkins) less stable. This pull request improves upon the previous situation.


toolslive:
ok.

PR #262, opened 2016-06-27
use all ascii in names of generated namespaces

### LINKED ISSUE #260: Namespace names in OSD backend or Cache backend are not UTF-8
Namespace names in OSD backend or Cache backend are not UTF-8. While not strictly required, UTF-8 is advised for JSON, and the python json parser seems to depend on this advise as it can't parse the JSON output returned by e.g. show-namespaces.

It would be nice if the counter could be added in some more readable way, e.g. `<prefix>_<counter>` where counter is `00000001`, `00000002`, and so on.

domsj:
fixes #260


toolslive:
ok.

PR #263, opened 2016-06-28
Rora

toolslive:
@redlicha please be gentle ;)


cpp/include/rdma_proxy_client.h
```diff
@@ -0,0 +1,50 @@
...
+
+*/
+
+#pragma once
+#include "generic_proxy_client.h"
+
+namespace alba {
+namespace proxy_client {
+
+class RDMAProxy_client : public GenericProxy_client {
```
domsj: Any reason you preferred inheritance over composition here?
In general the advice goes the other way ...
(Assuming it's still ok to discuss design)

cpp/src/lib/manifest_cache.cc
```diff
@@ -0,0 +1,49 @@
...
+  _cache_mutex.lock();
+
+  auto it = _cache.find(key);
+  if (it != _cache.end()) {
+    _cache.erase(it);
+  }
+  _cache.emplace(std::make_pair(key, std::move(mfp)));
+
+  _cache_mutex.unlock();
+}
```
redlicha: I think the lookup / remove / insert can be simplified to `_cache[key] = std::move(mfp);`.
But taking a step back (and without any test data yet), this global lock could become quite popular / cause a lot of contention. In that case (and btw, what about cache invalidation?) another layer of indirection could help: a map or sth. similar for the namespaces and then per namespace another map for the manifests. Each level with its own locking (R/W locks) / the namespace map could possibly even be done without locking and atomic ops instead in the read path.

NB: RAII (std::lock_guard) would be better for locking instead of explicit lock / unlock

cpp/src/lib/manifest_cache.cc
```diff
@@ -0,0 +1,49 @@
...
+
+manifest_cache::iterator ManifestCache::find(strpair &key) {
+  return _cache.find(key);
+}
```
redlicha: No locking?

cpp/include/rdma_proxy_client.h
toolslive: 2 reasons:
- it's shorter.
- there's no need for extra flexibility in this case.

cpp/src/lib/osd_access.cc
```diff
@@ -0,0 +1,130 @@
...
+    ALBA_LOG(DEBUG, "osd:" << osd << " ip:" << ip
+                           << " port: " << backdoor_port);
+    int err = ctx_attr_set_transport(ctx_attr, transport_name.c_str(),
+                                     ip.c_str(), backdoor_port);
+    ALBA_LOG(DEBUG, "set_transport err:" << err);
+    assert(err == 0);
+    std::shared_ptr<gobjfs::xio::client_ctx> ctx = ctx_new(ctx_attr);
+    err = ctx_init(ctx);
+    ALBA_LOG(DEBUG, "ctx_init err:" << err);
+    assert(err == 0);
```
redlicha: While ctx_new et al don't throw exceptions (yet?) but use return codes these need to be dealt with differently, as the asserts might be compiled out in release builds aimed at performance and that could lead to use use of uninitialized data. For starters a simple macro that maintains the assert if NDEBUG is not set and otherwise throws some std::exception might be good enough to avoid this, in the long run gobjfs should be converted to throw exceptions.

cpp/src/lib/osd_access.cc
```diff
@@ -0,0 +1,130 @@
...
+    _osd_ctxs.emplace(osd, ctx);
+    ctx_it = _osd_ctxs.find(osd);
+  }
+  auto ctx = ctx_it->second;
+
+  std::vector<giocb *> iocb_vec;
+  std::vector<std::string> key_vec;
+  for (auto &slice : slices) {
+
+    giocb *iocb = (giocb *)malloc(sizeof(giocb));
```
redlicha: There's no need for the allocation (and also no need for malloc) inside the loop - it could be moved out and the explicit free'ing be avoided by either putting it on the stack (if we're certain it won't explode ;)) or by allocating the vector in one go - `std::vector<giocb>`.

ocaml/src/capabilities.ml
```diff
@@ -0,0 +1,54 @@
...
+as published by the Free Software Foundation, in version 3 as it comes
+in the <LICENSE.txt> file of the Open vStorage OSE distribution.
+
+Open vStorage is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY of any kind.
+ *)
+
+
+
+module OsdCapabilities = struct
```
domsj: Are these really osd capabilities or rather asd capabilities? (The todos in alba_osd.ml & kinetic_client.ml point towards asd capabilities?)
And is this concept actually needed?
I only expected some form of osd info to be sent back by the proxy OsdInfo call when we're talking about an actual 'rora'.
```
type rora_info = { ip : string; port : int; } (* note that port is the rora-port; maybe a few other things are needed too? *)
```

and then in the proxy protocol definition
```
| OsdInfo : (unit, osd_id * rora_info) Std.counted_list
```

ocaml/src/tools/alba_rora_stubs.c
```diff
@@ -0,0 +1,219 @@
...
+    tgt[1] = h1;
+}
+
+
+int _get_fnr_flen(char* returned_value, uint64_t *fnr, uint32_t *flen){
+    printf("_get_fnr_flen\n");fflush(stdout);
+    //
+    // 03 3f 57 45 d9 02 00 00 00 00 00 00 00 00 d7 f2 0a 00
+    int pos = 0;
+    printf("xxxx\n");fflush(stdout);
```
domsj: printf debugging (here and also further below)

ocaml/src/tools/alba_rora_stubs.c
```diff
@@ -0,0 +1,219 @@
...
+    uint8 checksum_type = returned_value[pos++];
+    printf("checksum_type:%i\n", checksum_type);fflush(stdout);
+    assert(checksum_type = 3);
```
domsj: does this mean rora won't work when a different checksum than crc32c is configured?

ocaml/src/tools/alba_rora_stubs.c
```diff
@@ -0,0 +1,219 @@
...
+    pos += 4;// crc
+    uint8 blob_type = returned_value[pos++];
+    printf("blob_type:%i\n", blob_type);fflush(stdout);
+    assert(blob_type = 2);//on Fs.
```
domsj: does this mean that rora won't work with small blobs (that are not written to the fs)?

ocaml/src/tools/alba_rora_stubs.c
toolslive: No. the printf & asserts here just mean I'm not finished yet. The pull request was opened to let @redlicha already review the c++ code.

ocaml/src/capabilities.ml
toolslive: Ironically, they were first called AsdCapabilities, but I changed my mind. I must admit the difference between Asd and Osd is no longer that clear to me. 
Osd started out as the common abstract interface that both Kinetic and Asds supported, but that's not really the case anymore.

The concept should allow us to remove the various tests we do to see if individual operations are supported or not( [for example here](https://github.com/openvstorage/alba/blob/master/ocaml /src/asd_client.ml#L134) )

domsj: If the idea is to remove the tests for availability of individual operation then to me these are asd specific and thus asd capabilities?

toolslive: why wouldn't Kinetic drives have capabilities too, or an alba osd for that matter?

cpp/src/lib/osd_info.cc
```diff
@@ -0,0 +1,88 @@
...
+template <> void from(message &m, proxy_protocol::OsdCapabilities &caps) {
+  uint32_t size;
+  from(m, size);
+  for (uint i = 0; i < size; i++) {
+    uint8_t version;
+    from(m, version);
+    assert(version == 1);
+    uint8_t tag;
+    from(m, tag);
+    switch (tag) { case 1: { }; break; // for now, we don't care
```
domsj: better put an `assert (tag = 3)` here then as otherwise you could get unexpected results?

domsj:
Which tests do we want for this?
- happy path (this you probably already tested)
- test whether fallback path works when using either encryption or compression
- some asds die during the test
- ...?

(We really need rdma capable jenkins slaves)

cpp/src/lib/osd_info.cc
toolslive: `case 1: ` means this osd can do a `MultiGet2`. I think. for now we really don't care.

toolslive:
- happy path is ok, and has some tests.
- fallback path works too. It suffices to do a download with the rora client for the default namespace.
  It's a bit problematic to verify that the proxy path is actually taken as there's no way to assert afterwards which path was taken unless we change the api to return that information.
- dying asds during test, after we have a connection is needed as I rely on gobjfs to do the right thing but we need to "trust but verify".
- we really do need rdma capably jenkins slaves, but xio works with tcp too.

ocaml/src/rora_server.ml
```diff
@@ -0,0 +1,50 @@
+open Foreign
```
domsj: missing license header

domsj:
ok

PR #273, opened 2016-07-05
Fix tests

toolslive:
- missing dependency for debian/rpm packages
- tests in everything else were failing
- add ovs apt repo in integration image to allow install of libboost1.57-all


domsj:
ok

domsj:
alba_centos7_verify docker image needs to be updated too?

toolslive:
indeed.

PR #275, opened 2016-07-07
Lru cache eviction

domsj:
Still has conflicts with master, will update it soon


domsj:
There should be a new version of ocaml-redis sometime soon, then we no longer need the CI modifications I made in this pull request

ocaml/src/asd_server.ml
```diff
@@ -1646,7 +1653,7 @@ let run_server
                  ~tcp_keepalive
                  ~tls:None protocol
        in
-       t :: threads
+       (wrap_t t "plain server") :: threads
```
toolslive: why is this needed?

ocaml/src/alba_eviction.ml
```diff
@@ -108,66 +108,106 @@ let do_eviction
...
+    key
+    0. max_float
+    ~limit:(0, 100)
+  >>= fun items ->
+
+  let items =
+    List.map
+      (fun r ->
+       match r with
+       | `Bulk (Some b) -> b
```
toolslive: Where is this created (I found only one occurrence of `Bulk) ?

ocaml/src/asd_server.ml
domsj: This change wasn't needed, I added it while debugging some problem, but in the end it turned out the asds weren't to blame.
Still thought the change could be useful so I left it in.

ocaml/src/alba_eviction.ml
domsj: The items are added to redis in https://github.com/openvstorage/alba/pull/275/files#diff-044afbbce530f0adfcecda0062438fbdR100 .
`Bulk is just the way the ocaml redis library returns the results from zrangebyscore.

toolslive:
ok.

PR #280, opened 2016-07-18
new gobjfs version

toolslive:
respect the number of cores in config.
(I'm also expecting another bugfix, so we might stall a bit)


domsj:
ok

PR #285, opened 2016-07-19
avoid reading outside the message

toolslive:
could for example happen if server sends bogus data


domsj:
Even with this change it's still possible that the deserialisation will read a bit outside of the buffer, right?
But I guess it should now fail with a proper exception for the specific case that you saw...
LGTM

toolslive:
how could we read outside the message ?
all reading starts by fetching the `char*` like this
`const char *ib = m.current();`
Then we look in at `ib` and beyond.

domsj:
Suppose a message is 1 byte long, and next we try to read an int32 from it. If I'm not mistaken that will result in a read outside the buffer...

domsj:
Ah, so inside the skip method we could detect that we've read sth from outside the buffer

toolslive:
regarding the 1 byte: you're right.
regarding the skip, it might be too late (already trapped) .

domsj:
You're right that it could already be too late
But still I think it could be a useful improvement. Combined with changing the order of reading from the buffer and doing the skip it's not too late anymore.

e.g. change
```
template <> void from(message &m, uint64_t &i) {
  const char *ib = m.current();
  i = *((uint64_t *)ib);
  m.skip(8);
}
```
to
```
template <> void from(message &m, uint64_t &i) {
  const char *ib = m.current();
  m.skip(8);
  i = *((uint64_t *)ib);
}
```

domsj:
ok, looks good!

PR #286, opened 2016-07-19
Configurable transport for rora

domsj:
This hasn't been tested. Ideally we would have rdma capable jenkins slaves, then I would add an extra test suite for the possible transport combinations...
I will try to do a manual test tomorrow (will probably require some changes to setup.ml).
Opening this pull request now so it can already be reviewed.
The new code will also have to be integrated by @redlicha once it's been approved/merged...
Note that old clients speaking with newer proxies will currently fail on some assert .. I can make some more changes to avoid such a scenario if desired.


cpp/src/lib/osd_info.cc
```diff
@@ -83,13 +91,19 @@ std::ostream &operator<<(std::ostream &os, const OsdInfo &info) {
 }
 
 std::ostream &operator<<(std::ostream &os, const OsdCapabilities &caps) {
-  os << "OsdCapabilities( port= ";
+  os << "OsdCapabilities( rora_port= ";
+
+  if (boost::none == caps.rora_port) {
```
toolslive: add an `operator<<` for `boost::optional` to `stuff.h` ? (alternative: there's something available in boost as well, if you add the right `#include` but it doesn't show it's an optional anymore)

toolslive:
ok, once you've manually verified it.

Issue #259, opened 2016-06-24
Exception in populate_osds_info_cache: (Failure "attempting to use a resource from a pool which is being finalized");


kinvaris:
During our multidatacenter we encountered this in a proxy:
```
2016-06-24 15:38:40 118093 +0200 - perf-roub-03 - 40723/0 - alba/proxy - 945065 - info - Exception in populate_osds_info_cache: (Failure "attempting to use a resource from a pool which is being finalized"); backtrace:; Raised at file "map.ml", line 117, characters 16-25; Called from file "src/core/lwt.ml", line 161, characters 4-40
2016-06-24 15:38:40 126257 +0200 - perf-roub-03 - 40723/0 - alba/proxy - 945066 - info - Exception in populate_osds_info_cache: (Failure "attempting to use a resource from a pool which is being finalized"); backtrace:; Raised at file "map.ml", line 117, characters 16-25; Called from file "src/unix/lwt_engine.ml", line 270, characters 8-36
```
This error was filling the logfile of the proxy very quickly.

domsj:
Fixed in #289, should be in the next alba release

PR #289, opened 2016-07-20
#259

domsj:
fixes the abundance of log messages mentioned in #259 and considerably reduces the number of connections set up towards the asds


toolslive:
ok.

Issue #233, opened 2016-06-08
Alba osd: refresh abmcfg + provide an implementation for its disk usage


domsj:

domsj:
the same applies for 'proxy osd'

PR #291, opened 2016-07-20
Refresh alba osd abm cfg

domsj:


domsj:
This partially addresses #233

ocaml/src/maintenance.ml
```diff
@@ -1999,8 +2001,68 @@ class client ?(retry_timeout = 60.)
...
+           (fun (osd_id, osd_info, alba_cfg) ->
+            let alba_osd_arakoon_cfg = alba_cfg.Nsm_model.OsdInfo.cfg in
+            Lwt.catch
+              (fun () ->
+               Albamgr_client.with_client'
+                 alba_osd_arakoon_cfg
+                 ~tls_config
+                 ~tcp_keepalive
+                 (fun client ->
+                  Lwt.catch
```
toolslive: maybe only use 1 level of  Lwt.catch here (and log in the outer one) ?

toolslive:
ok.

Issue #297, opened 2016-07-26
rora: client's manifest cache should also be filled on partial read via proxy.


toolslive:

domsj:
fixed by #299

PR #299, opened 2016-07-26
Partial read with manifest

toolslive:


domsj:
this should fix https://github.com/openvstorage/alba/issues/297, will review/merge soon

cpp/src/examples/test_client.cc
```diff
@@ -90,98 +97,92 @@ void proxy_get_version(const string &host, const string &port,
...
 
 using namespace std::chrono;
+void _bench_one_client(std::unique_ptr<Proxy_client> client,
+                       std::shared_ptr<alba::statistics::Statistics> stats_p,
+                       const string &namespace_, const string &object_name,
+                       const string &file_name, const int n) {
+
+    try{
+        const alba::Checksum *checksum = nullptr;
+        alba::statistics::Statistics &stats = *stats_p;
```
domsj: statistics are tracked with this object, but it's never used to print/produce output?

cpp/src/lib/manifest_cache.cc
```diff
@@ -31,35 +31,87 @@ void ManifestCache::set_capacity(size_t capacity) {
...
+  manifest_cache &manifest_cache = *mcp;
+  std::mutex &m = *mp;
+  {
+    std::lock_guard<std::mutex> lock(m);
+    auto it2 = manifest_cache.find(object_name);
+    if (it2 != manifest_cache.end()) {
+      manifest_cache.erase(it2);
+    }
+
+    if (manifest_cache.size() > _manifest_cache_capacity) {
```
domsj: So this means the manifest cache size is now per namespace instead of globally.
That's fine by me, but we have to make sure to communicate this to users of the client.

cpp/src/lib/manifest_cache.cc
```diff
@@ -31,35 +31,87 @@ void ManifestCache::set_capacity(size_t capacity) {
...
+      auto it_victim = manifest_cache.begin();
```
domsj: I know this hasn't changed in this pull request, but I'm wondering if this victim selection could lead to consistently getting and evicting the manifest of the same object(s). We may want to improve this in the future?

cpp/src/lib/rora_proxy_client.cc
```diff
@@ -221,20 +229,38 @@ int RoraProxy_client::_short_path_many(
...
+void _process(std::vector<object_info> &object_infos,
+              const std::string &namespace_) {
+
+  ALBA_LOG(DEBUG, "_process : " << object_infos.size());
+  for (auto &object_info : object_infos) {
+    using alba::stuff::operator<<;
+    // ALBA_LOG(DEBUG, "_procesing object_info:" << object_info);
+
+    const std::string &object_name = std::get<0>(object_info);
+    // auto& future = std::get<1>(object_info);
```
domsj: Should check here whether future has the expected value. If not it should not be added to the manifest cache.

cpp/src/lib/rora_proxy_client.cc
```diff
@@ -221,20 +229,38 @@ int RoraProxy_client::_short_path_many(
...
     std::vector<short_path_entry> short_path;
     for (auto &object_slices : slices) {
       auto object_name = object_slices.object_name;
-      auto key = strpair(namespace_, object_name);
       auto &cache = ManifestCache::getInstance();
-      auto mfp = cache.find(key);
+      auto mfp = cache.find(namespace_, object_name);
       if (nullptr != mfp &&
           compressor_t::NO_COMPRESSION == mfp->compression->get_compressor() &&
           encryption_t::NO_ENCRYPTION == mfp->encrypt_info->get_encryption()) {
```
domsj: redundant check as this (no compression, no encryption) was already checked before adding the manifest to the cache

ocaml/src/proxy_server.ml
```diff
@@ -153,13 +153,34 @@ let read_objects_slices
...
         >>= function
         | None -> Protocol.Error.failwith Protocol.Error.ObjectDoesNotExist
-        | Some (_mf, mf_src) -> Lwt.return mf_src
+        | Some (mf, namespace_id, mf_source) -> Lwt.return (object_name, mf, namespace_id, mf_source)
        )
        objects_slices
-     >>= fun mf_sources ->
+     >>= fun n_mf_src_s ->
+
+     let objects_infos =
```
domsj: This whole transformation into object_infos is a bit weird (possibly correct but hard to follow).
I expected something more along the lines of
```
List.map
  (fun (name, mf, namespace_id, _) -> (name, "level0", (mf, namespace_id)))
  n_mf_src_s
```
?

ocaml/src/proxy_server.ml
```diff
@@ -186,7 +207,15 @@ let render_request_args: type i o. (i,o) Protocol.request -> i -> Bytes.t =
                            namespace
                            ([%show : (object_name *
                                         (offset * length) list) list ]
-                           objects_slices)
+                              objects_slices)
+  | ReadObjectsSlices2 -> fun (namespace, objects_slices, consistent_read) ->
```
domsj: extract render_read_object_slices

domsj:
Already addressed a few of my own comments so it's good to merge. Would like to discuss some of the others with @toolslive when he's back.

Issue #306, opened 2016-08-04
Make help of  alba show-namespaces more clear


wimpers:
Currently the help of  show-namespaces is quite limited: ' information about a range of namespaces'. Add more information about how one should read the output of this command.

domsj:
fixed in #308

domsj:
https://github.com/openvstorage/alba/pull/308#issuecomment-237784555

PR #308, opened 2016-08-05
some help improvements as requested in #306

domsj:


wimpers:
@domsj , can't we add an example output and explain the output?

domsj:
We can ... reopened your issue for that

Issue #318, opened 2016-08-11
RDMA Proxy client's timeout shouldn't be based on gettimeofday


redlicha:
Discovered via code inspection while debugging https://github.com/openvstorage/alba/issues/317 :
```
double _stamp_ms() {
  struct timeval tp;
  gettimeofday(&tp, NULL);
  double t0 = 1000 * tp.tv_sec + (double)tp.tv_usec / 1e3;
  return t0;
}
```

`gettimeofday` returns the wall clock time - better use a steady clock, e.g. `boost::chrono::steady_clock` - http://www.boost.org/doc/libs/1_61_0/doc/html/chrono/reference.html#chrono.reference.cpp0x.system_clocks_hpp.steady_clock , e.g. (pseudo code!):

```
auto end = boost::chrono::steady_clock::now() + timeout_duration;
do
{
    // ... work
}
while (boost::chrono::steady_clock::now() < end)
```

domsj:
fixed in #325

PR #325, opened 2016-08-12
Rdma proxy client fixes

### LINKED ISSUE #317: RDMA Proxy client: infinite loop reading the proxy response
Observed during tests:
```
void RDMAProxy_client::_really_read(char *buf, const int len) {
  // ... snip ...
  while (todo > 0 && _request_time_left > 0) {
    double t0 = _stamp_ms();

    // ... snip ... 
    int rc = rpoll(&pollfd, nfds, _request_time_left);
    
    // ... snip ...
    read = rrecv(_socket, &buf[off], todo, flags);
    if (read < 0) {
      throw proxy_exception(read, _build_msg("really_read.rrecv"));
    }

     // ... snip ...
    double t1 = _stamp_ms();
    double delta = t1 - t0;
    _request_time_left = _request_time_left - (int)delta;
  }
}
```

The case `read == 0` is not handled correctly:
> The return value  will  be  0 when the peer has performed an orderly shutdown

(from `recv(2)`).

Together with the rounding caused by `_stamp_ms()` and the `double` -> `int` conversion, this can cause `_request_time_left` to stay at its current value forever.

domsj:
Should fix #317 & #318 

@redlicha can you review?

(will test on 154.231)

### DIFF
--- a/cpp/src/lib/rdma_proxy_client.cc
+++ b/cpp/src/lib/rdma_proxy_client.cc
@@ -27,13 +27,6 @@ namespace proxy_client {
 using std::string;
 using llio::message;
 
-double _stamp_ms() {
-  struct timeval tp;
-  gettimeofday(&tp, NULL);
-  double t0 = 1000 * tp.tv_sec + (double)tp.tv_usec / 1e3;
-  return t0;
-}
-
 std::string _build_msg(const std::string &prefix) {
   int _errno = errno;
   std::ostringstream ss;
@@ -42,8 +35,11 @@ std::string _build_msg(const std::string &prefix) {
 }
 
 void RDMAProxy_client::_really_write(const char *buf, const int len) {
-  if (len <= 0) {
+  if (len == 0) {
     return;
+  }
+  if (len < 0) {
+    throw proxy_exception(len, _build_msg("really_write negative length"));
   }
 
   int flags = 0;
@@ -92,8 +88,11 @@ void RDMAProxy_client::_really_write(const char *buf, const int len) {
 }
 
 void RDMAProxy_client::_really_read(char *buf, const int len) {
-  if (len <= 0) {
+  if (len == 0) {
     return;
+  }
+  if (len < 0) {
+    throw proxy_exception(len, _build_msg("really_read negative length"));
   }
 
   int flags = 0;
--- a/cpp/src/lib/rdma_proxy_client.cc
+++ b/cpp/src/lib/rdma_proxy_client.cc
@@ -19,7 +19,6 @@
 #include "rdma_proxy_client.h"
 #include <rdma/rsocket.h>
 #include <boost/lexical_cast.hpp>
-#include <chrono>
 
 namespace alba {
 namespace proxy_client {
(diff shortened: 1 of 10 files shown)


cpp/src/lib/rdma_proxy_client.cc
```diff
@@ -41,64 +42,90 @@ std::string _build_msg(const std::string &prefix) {
 }
 
 void RDMAProxy_client::_really_write(const char *buf, const int len) {
+  if (len <= 0) {
+    return;
```
redlicha: Meta question as this is not something newly introduced: is `len < 0` something that the code should tolerate, or rather something to complain about via `assert` / exception, or to be ruled out altogether with an unsigned typed?

cpp/src/lib/rdma_proxy_client.cc
```diff
@@ -41,64 +42,90 @@ std::string _build_msg(const std::string &prefix) {
...
+
   int flags = 0;
   int sent;
   int todo = len;
   int off = 0;
   int nfds = 1;
 
-  while (todo > 0 && _request_time_left > 0) {
-    double t0 = _stamp_ms();
-    ALBA_LOG(DEBUG, "todo=" << todo
```
redlicha: `_stamp_ms()` is unused now AFAICS and can be removed.

cpp/src/lib/rdma_proxy_client.cc
```diff
@@ -41,64 +42,90 @@ std::string _build_msg(const std::string &prefix) {
...
-    double t1 = _stamp_ms();
-    double delta = t1 - t0;
-    _request_time_left = _request_time_left - (int)delta;
-  }
+  } while (todo > 0 && time_remaining.count() > 0);
+
+  time_remaining = end - std::chrono::steady_clock::now();
+  _request_time_left -= std::chrono::duration_cast<std::chrono::milliseconds>(
+                            time_remaining).count();
 }
```
redlicha: I think replacing `RDMAProxy_client`'s `int request_time_left /* ms */` member with a `chrono::steady_clock::time_point _deadline` member could simplify all of this by avoiding a few conversions between `int`, `time_point` and `duration`, e.g.
```
while (// there is data to be sent)
{
    const auto now = chrono::steady_clock::now();
    if (now >= _deadline)
    {
          // throw timeout exception
    }
    const auto time_left = _deadline - now;
    poll(..., duration_cast<milliseconds>(time_left).count());
    // handle poll errors
    send(...);
    // handle send errors
}
````

cpp/src/lib/rdma_proxy_client.cc
domsj: indeed forgot to remove this

cpp/src/lib/rdma_proxy_client.cc
domsj: ok

cpp/src/lib/rdma_proxy_client.cc
domsj: ok, will throw an exception there

toolslive:
fine for me.

Issue #321, opened 2016-08-11
c++ proxy client "reading outside of message"


domsj:
```
2016-08-10 17:00:00 153126 +0200 - cmp03 - 22174/0x00007eff2fe7b700 - volumedriverfs/AlbaProxyClient - 00000000000056fd -  warning - logMessage: WARNING: _pos:0 + 4 > 0
2016-08-10 17:00:00 153286 +0200 - cmp03 - 22174/0x00007eff2fe7b700 - volumedriverfs/AlbaConnection - 00000000000056fe -  error - convert_exceptions_: write object: caught exception: reading outside of message
```

could be related to #320 ?

domsj:
Fixed by #331

Issue #320, opened 2016-08-11
c++ proxy client "unexpecteded Manifest version"


domsj:
```
2016-08-10 15:47:47 029128 +0200 - cmp03 - 22174/0x00007efe0132e700 - volumedriverfs/AlbaConnection - 0000000000004500 -  error - convert_exceptions_: write object: caught exception: unexpecteded Manifest version
```

domsj:
Fixed by #331

PR #331, opened 2016-08-17
take eofbit into account too when evaluating state of an input/output stream

domsj:
This probably explains #320 and #321


toolslive:
ok.

PR #332, opened 2016-08-18
Rora server fixes

toolslive:
and a bit more


ocaml/src/cli_common.ml
```diff
@@ -217,6 +217,11 @@ let verbose =
        & flag
        & info ["verbose"] ~doc:"more output on cli")
 
+let unescape =
```
domsj: this is already defined in cli_osd too, maybe clean up that one?

domsj:
ok

Issue #319, opened 2016-08-11
Proxy: (Invalid_argument "index out of bounds")


domsj:
```
2016-08-11 09:24:04 300931 +0200 - cmp03 - 15924/0 - alba/proxy - 7897 - info - server: exception occurred in client connection (172.19.12.142,49799): (Invalid_argument "index out of bounds")
```

this is triggered during a write call from the voldrv:
```
2016-08-11 09:24:03 211011 +0200 - cmp03 - 15928/0x00007f0e56330700 - volumedriverfs/BackendConnectionInterfaceLogger - 000000000001038e -  error - ~Logger: Exiting write for 01756be4-f8ae-45b0-80a9-4ee206b070ea sco_access_data with exception
```

domsj:
I think we haven't seen this any more. We added some more logging related to this.
Can the issue be closed?

wimpers:
Please re-open if it occurs again

PR #333, opened 2016-08-18
show where (index out of bounds) exceptions come from

domsj:
related to #319 
should show if the exception comes from a malformed request or somewhere during processing of the request


toolslive:
ok.

PR #340, opened 2016-08-19
asds can be checked for consistency

toolslive:


ocaml/src/asd_server.ml
```diff
@@ -682,6 +682,87 @@ let execute_query : type req res.
...
+       in
+       let checksum_ok cv =
+         let (checksum, _) = cv in
+         let algo = Checksum.algo_of checksum in
+         Lwt.catch
+           (fun () ->
+             Value.get_blob_from_value dir_info cv >>= fun blob ->
+             let hash = Hashes.make_hash algo in
+             let open Slice in
+             let () = hash # update_substring blob.buf blob.offset blob.length in
```
domsj: I think we want to calculate the hashes of big blobs on other threads...
So files should be read into a Lwt_bytes.t (avoid reading to Bytes.t and then make another copy...) and then use `hash # update_lwt_bytes_detached`

ocaml/src/asd_server.ml
```diff
@@ -682,6 +682,87 @@ let execute_query : type req res.
...
+                 then
+                   (cnt+1, Some key, (key, check_results) :: acc)
+                 else
+                   (cnt, Some key, acc)
+               in
+               Lwt.return acc'
+             ) (0, None, []) items
+           >>= fun (cnt', last, r) ->
+           let r' = List.rev r in
+           let cost = max * 1000 in (* TODO: better estimate *)
```
domsj: The cost should probably also depend on the type of check that was performed?
Verifying a checksum (which requires reading the entire blob) is a lot more costly than just verifying the file still exists.

ocaml/src/cli_asd.ml
```diff
@@ -370,6 +370,83 @@ let asd_range_cmd =
   in
   asd_range_t, info
 
+let asd_range_validate hosts port transport tls_config asd_id first
+                       verbose unescape verify_checksum follow show_all =
+  let finc = true
+  and last = None in
```
domsj: when passing in --follow you probably also want to be able to specify last?
or maybe allow --follow within a prefix?
(that way you could manually trigger verify of the keys for a certain namespace)

ocaml/src/cli_asd.ml
```diff
@@ -370,6 +370,83 @@ let asd_range_cmd =
...
+          (fun (slice, checks) ->
+            Lwt_io.printlf "%s %s" (Slice.show slice) ([%show: (string * bool) list] checks)
+          )
+          items
+        >>= fun () ->
+        match next_key_o with
+        | None -> Lwt.return_unit
+        | Some key ->
+           let first' = Slice.get_string key in
+           loop first' false
```
domsj: I think you forgot to check `follow` here before looping

ocaml/src/cli_asd.ml
```diff
@@ -370,6 +370,83 @@ let asd_range_cmd =
...
+      in
+      loop first' finc
+    )
+
+let asd_range_validate_cmd =
+  let first =
```
domsj: this is already defined in cli_common.ml

domsj:
This could be used in verify.ml to avoid transferring the blob over the network.
There's enough in this pull request already so we probably want to create a follow up ticket for that?

ocaml/src/cli_asd.ml
toolslive: last would be interesting. will add.

ocaml/src/cli_asd.ml
toolslive: indeed

ocaml/src/asd_server.ml
```diff
@@ -410,6 +410,109 @@ module Net_fd = struct
...
+           DirectoryInfo.with_blob_fd
+             dir_info fnr
+             (fun fd ->
+               Buffer_pool.with_buffer
+                 Buffer_pool.default_buffer_pool
+                 (fun buffer ->
+                   let reader = Lwt_bytes.read fd in
+                   let writer = hash # update_lwt_bytes_detached
+                   in
+                   Lwt_extra2.copy_using reader writer size buffer)
```
domsj: This will read and hash the entire blob in slices of 4KB (due to the default buffer pool size).
For efficiency it would be better to read bigger slices, e.g. of 1MB.

ocaml/src/asd_server.ml
```diff
@@ -410,6 +410,109 @@ module Net_fd = struct
...
+            then
+              (cnt+1, Some key, (key, check_results) :: acc)
+            else
+              (cnt, Some key, acc)
+          in
+          Lwt.return acc'
+        ) (0, None, []) items
+      >>= fun (cnt', last, r) ->
+      let r' = List.rev r in
+      let factor = if verify_checksum then 1000 else 100 in
```
domsj: When the checksum was verified the cost should be proportional to the size of the blob?

toolslive: doubt it. for 16KB that's in rocksdb it's almost free. For 16KB+1B that's on the FS it's suddenly way more expensive. In all, it's continuous nor linear, and we're comparing apples and oranges anyway.

domsj: Ok, only part of the cost is proportional ... I was thinking of reading+hashing 1MB versus 4MB blobs.

domsj:
Ok. Created https://github.com/openvstorage/alba/issues/347 as follow up item for my earlier remark.

PR #342, opened 2016-08-22
per namespace LRU for manifest cache

toolslive:
maybe @redlicha wants to have a look too.

### DIFF
--- a/cpp/src/lib/lru_cache.h
+++ b/cpp/src/lib/lru_cache.h
@@ -1,0 +1,106 @@
+// Copyright (C) 2016 iNuron NV
+//
+// This file is part of Open vStorage Open Source Edition (OSE),
+// as available from
+//
+//      http://www.openvstorage.org and
+//      http://www.openvstorage.com.
+//
+// This file is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


cpp/src/lib/lru_cache.h
```diff
@@ -0,0 +1,106 @@
...
+//
+// This file is free software; you can redistribute it and/or modify it
+// under the terms of the GNU Affero General Public License v3 (GNU AGPLv3)
+// as published by the Free Software Foundation, in version 3 as it comes in
+// the LICENSE.txt file of the Open vStorage OSE distribution.
+// Open vStorage is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY of any kind.
+
+#ifndef LRU_CACHE_TOO_H_
+#define LRU_CACHE_TOO_H_
```
redlicha: Please change the include guard's name (prefix with ALBA_?) to prevent collisions with voldrv's version (should it end up being exposed to it via a public header).

domsj: or use `#pragma once` as we do in other places?

domsj:
looks good to me

Issue #208, opened 2016-05-26
Repair on read


domsj:
When a (partial) object read can't get a copy of a certain fragment* we have to regenerate it from the other fragments of the chunk.
This incurs a (big) cost (extra IO to fetch the other fragments and some cpu), which will have to be paid each time until the fragment gets repaired by the maintenance process.
Ok, reality is better than this: if the proxy is configured with a fragment cache the price only has to be paid when the fragment can not be found in the cache.
Do note that all flash alba's will (usually) be configured to not use a fragment cache (so that they can do partial reads from the asds), so there this situation is costly.

We can improve this behaviour (only pay the price once) by doing a repair on read - that is store the fragment on an available osd while we have it in memory anyway.

*: This could be due to several reasons: e.g. 1) the fragment current just doesn't exist (location None for it in the manifest) or 2) it somehow got corrupted or 3) the osd doesn't have the fragment (due to some bug?) or 4) the osd is unreachable

hofkensj:
@wimpers do we have an ETA on this one.  We need it for the Sberbank tests.

wimpers:
@hofkensj, @domsj is currently working on it. Expected ETA before end of next week (exact date will depend on the bugs popping up from the RORA tests). Basically the team is trying to release it asap.

hofkensj:
I assume this will be in Fargo right .. not unstable.

domsj:
Repair on read has been implemented for the partial read path, where the need was the most pressing.
This can also be done for the full object read path, but that could use some other optimisation too (now it always tries to fetch all fragments which isn't really needed).
So should we lower the priority of this ticket, or close it and create another one about optimising the (full) object read path?

PR #373, opened 2016-09-15
Refresh manifest on missing fragment

### LINKED ISSUE #371: Manifest cache does not get properly refreshed after maintenance agent action
Observed during rora performance testing
@domsj: pls update with more clear description - thx

domsj:
This fixes #371.

As a next step (building on these changes) I plan on implementing https://github.com/openvstorage/alba/issues/208 for the partial read path (that's where it's most important).

### DIFF
--- a/ocaml/src/alba_client_download_slices.ml
+++ b/ocaml/src/alba_client_download_slices.ml
@@ -23,6 +23,9 @@ open Slice
 open Lwt_bytes2
 open Lwt.Infix
 
+exception Unreachable_fragment of { chunk_id : Nsm_model.chunk_id;
+                                    fragment_id : Nsm_model.fragment_id; }
+
 let blit_from_fragment fragment_data fragment_intersections =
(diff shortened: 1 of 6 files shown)


ocaml/src/alba_client_download_slices.ml
```diff
@@ -0,0 +1,533 @@
...
+          ~location
+          ~chunk_id ~fragment_id
+          ~replication
+          ~fragment_checksum
+          decompress
+          ~encryption
+          ~cache_on_read
+          fragment_cache
+          bad_fragment_callback)
+      (fun exn -> Lwt.fail (Unreachable_fragment { chunk_id; fragment_id; }))
```
toolslive: Do we care about the original exception ?

domsj: added more logging

toolslive:
ok.

PR #387, opened 2016-09-26
Proxy sequences

domsj:


cpp/include/proxy_client.h
```diff
@@ -98,6 +100,11 @@ class Proxy_client {
   get_object_info(const std::string &namespace_, const std::string &object_name,
                   const consistent_read, const should_cache) = 0;
 
+  virtual void apply_sequence(const std::string &namespace_,
+                              const write_barrier,
+                              const std::vector<std::shared_ptr<sequences::Assert>> &,
+                              const std::vector<std::shared_ptr<sequences::Update>> &) = 0;
+
   /* invalidate_cache influences the result of read requests issued with
```
redlicha: While that interface is definitely sufficient, WDYT about a convenience layer along those lines on top of this (the actual details of the `Assert` and `Update` interfaces can then be hidden):

```
class Assert;
class Update;

BOOLEAN_ENUM(ObjectExists);

class Sequence
{
public:
    using Updates = std::vector<std::unique_ptr<Update>>;
    using Asserts = std::vector<std::unique_ptr<Assert>>;

    Sequence(size_t assert_size_hint = 0, size_t update_size_hint = 0); // hints to reserve vector space
    
    Sequence& add_assert(const std::string&, ObjectExists);
          
    Sequence& add_assert(const std::string&, const std::string& id);

    Sequence& add_assert(const std::string&, const CheckSum&);

    Sequence& add_upload(const std::string&, const std::string& path, const CheckSum*);

    Sequence& add_delete(const std::string&);

private:
    Updates updates_;
    Asserts asserts_;
};
```

and then consequently

```
virtual void apply_sequence(const std::string&, const write_barrier, const Sequence&);
```

Usage example:
```
const auto seq = Sequence().add_assert("foo", ObjectExists::T).add_upload("bar", "/path/to/bar", nullptr);
client->apply_sequence("nspace", write_barrier::T, seq);
```

cpp/include/proxy_protocol.h
```diff
@@ -151,6 +152,13 @@ void read_read_objects_slices2_response(message &m, Status &status,
                                         const std::vector<ObjectSlices> &dest,
                                         std::vector<object_info> &object_infos);
 
+void write_apply_sequence_request(
+    message_builder &mb, const string &namespace_, const bool write_barrier,
+    const std::vector<std::shared_ptr<alba::proxy_client::sequences::Assert>> asserts,
+    const std::vector<std::shared_ptr<alba::proxy_client::sequences::Update>> updates);
+
```
redlicha: Creates copies, should be `const std::vector<....>&` for both asserts and updates.

cpp/include/proxy_sequences.h
```diff
@@ -0,0 +1,133 @@
...
+class Assert {
+public:
+  virtual ~Assert(){};
+  virtual void to(llio::message_builder &) const = 0;
+};
+
+class AssertObjectExists : public Assert {
+public:
+  AssertObjectExists(std::string name) : _name(name){};
+
```
redlicha: Throughout this file strings are copied, better use `const std::string&`.

cpp/include/proxy_sequences.h
```diff
@@ -0,0 +1,133 @@
...
+  std::string _name;
+};
+
+class AssertObjectDoesNotExist : public Assert {
+public:
+  AssertObjectDoesNotExist(std::string name) : _name(name){};
+
+  void to(llio::message_builder &mb) const {
+    mb.add_type(2);
```
redlicha: NB: not an issue, but you might want to consider using C++-11's `override` and `final` specifiers for virtual methods to give the compiler extra hints.

cpp/include/proxy_protocol.h
domsj: ok

cpp/include/proxy_sequences.h
domsj: ok

cpp/Makefile
```diff
@@ -31,6 +31,7 @@ LIBS_exec = -L/usr/local/lib \
 
 _OBJECTS = stuff.o manifest.o alba_logger.o \
            proxy_protocol.o llio.o checksum.o \
+           proxy_sequences.o \
            io.o proxy_client.o generic_proxy_client.o \
```
domsj: ok

cpp/include/proxy_client.h
domsj: added, please have a look

cpp/include/proxy_sequences.h
```diff
@@ -129,6 +131,53 @@ class UpdateDeleteObject final : public Update {
...
   std::string _name;
 };
+
+BOOLEAN_ENUM(ObjectExists);
+
+class Sequence {
+public:
+  Sequence(size_t assert_size_hint = 0, size_t update_size_hint = 0)
+      : _asserts(assert_size_hint), _updates(update_size_hint) {}
+
```
redlicha: The size hints need to be used with the respective vector's `reserve` method - passing them to the constructor will lead to `size_hint` `nullptr`s. being added as the first elements, and `add_assert` and friends then push new entries behind these `nullptr`s.

cpp/src/lib/rora_proxy_client.h
```diff
@@ -85,6 +85,10 @@ class RoraProxy_client : public Proxy_client {
                  const std::vector<std::shared_ptr<sequences::Update>> &,
                  std::vector<proxy_protocol::object_info> &);
 
+  virtual void apply_sequence(const std::string &namespace_,
+                              const write_barrier, const sequences::Sequence &,
+                              std::vector<proxy_protocol::object_info> &);
+
```
redlicha: There should be no need for this / the implementation in `rora_proxy_client.cc` - the method in `generic_proxy_client` (which need not be `virtual`) should work fine?

cpp/include/proxy_sequences.h
domsj: updated + added a test now that uses Sequence

cpp/src/lib/rora_proxy_client.h
domsj: ok

toolslive:
ok.
(somehow I expected the tup files to be updated too)

PR #390, opened 2016-10-03
Rora performance

toolslive:
- null io
- io patterns (fixed, random, stride)
- small improvement to osd_access to let only 1 client do the fetching
- remove some high frequency log statements

### DIFF
--- a/cpp/src/lib/osd_access.cc
+++ b/cpp/src/lib/osd_access.cc
@@ -61,14 +61,38 @@ void OsdAccess::_remove_ctx(osd_t osd) {
   _osd_ctxs.erase(osd);
 }
 
-void OsdAccess::update(std::vector<std::pair<osd_t, info_caps>> &infos) {
-  ALBA_LOG(DEBUG, "OsdAccess::update");
-  std::lock_guard<std::mutex> lock(_osd_infos_mutex);
 
-  _osd_infos.clear();
-  for (auto &p : infos) {
-    _osd_infos.emplace(p.first, std::move(p.second));
-  }
+void OsdAccess::update(Proxy_client& client) {
+    if(!_filling){
+        ALBA_LOG(INFO, "OsdAccess::update:: filling up");
+        {
+            std::lock_guard<std::mutex> f_lock(_filling_mutex);
+            _filling = true;
+        }
+
+        try{
+            std::lock_guard<std::mutex> lock(_osd_infos_mutex);
+            std::vector<std::pair<osd_t, info_caps>> infos;
+            client.osd_info(infos);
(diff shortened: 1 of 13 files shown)


cpp/src/lib/osd_access.cc
```diff
@@ -61,14 +61,38 @@ void OsdAccess::_remove_ctx(osd_t osd) {
...
 
-  _osd_infos.clear();
-  for (auto &p : infos) {
-    _osd_infos.emplace(p.first, std::move(p.second));
-  }
+void OsdAccess::update(Proxy_client& client) {
+    if(!_filling){
+        ALBA_LOG(INFO, "OsdAccess::update:: filling up");
+        {
+            std::lock_guard<std::mutex> f_lock(_filling_mutex);
```
domsj: don't you want/need to hold the lock for the entire duration of the `if(!_filling){` block?

cpp/src/lib/osd_access.cc
```diff
@@ -61,14 +61,38 @@ void OsdAccess::_remove_ctx(osd_t osd) {
...
-void OsdAccess::update(std::vector<std::pair<osd_t, info_caps>> &infos) {
-  ALBA_LOG(DEBUG, "OsdAccess::update");
-  std::lock_guard<std::mutex> lock(_osd_infos_mutex);
 
-  _osd_infos.clear();
-  for (auto &p : infos) {
-    _osd_infos.emplace(p.first, std::move(p.second));
-  }
+void OsdAccess::update(Proxy_client& client) {
+    if(!_filling){
```
domsj: you may need to specify that you want to do a 'volatile' read here?
(that's the c# terminology, in c++ the stuff in std::memory_order might be related)
@redlicha can you review this part?

nit: please do a clang-format on this file.

domsj: It's possible that 2 clients will enter this block simultaneously and still perform too much work.
This is typically solved by the following pattern (double checked locking):
```
if (need_to_do_sth) {
  take lock;
  if (need_to_do_sth) {
    do_thing();
    need_to_do_sth = false;
  } else {
    ...
  }
  release lock;
} else {
  ...
}

domsj: just remembered that @redlicha is on pto. @cnanakos can you review please?

toolslive: ok, double checked locking combined with std::atomic it will be.

cnanakos:
You can directly use atomic CAS but this is fine also.

toolslive:
master has plenty of formatting issues, which should be fixed by the latest format here.

PR #393, opened 2016-10-03
url changed

toolslive:


domsj:
If we're changing the url it might make more sense to get a build from this job: http://10.100.129.100:8080/job/volumedriver-no-dedup-dev-release-ubuntu-14.04/ ? A new (recent enough) build should first be pinned there though.
I'm guessing the other job will probably not live long anymore. (And if it were to keep on living the uris would break on it's next build.)

PR #397, opened 2016-10-11
Rora reads from the fragment cache

domsj:
still to do:
- expose statistic about fast vs slow path; assert on them in the tests
- ping @redlicha to make new voldrv testers

but it's a lot of changes so please have a look already


cpp/src/lib/rora_proxy_client.cc
```diff
@@ -154,53 +166,154 @@ void RoraProxy_client::_maybe_update_osd_infos(
...
+}
+
+void _resolve_slice_one_level(std::vector<std::pair<byte *, Location>> &results,
+                              ManifestWithNamespaceId &manifest,
+                              uint64_t offset, uint32_t length, byte *target) {
+  while (length > 0) {
+    Location l = get_location(manifest, offset, length);
+    length -= l.length;
+    offset += l.length;
+    results.push_back(std::move(std::pair<byte *, Location>(target, l)));
```
toolslive: results in a copy ? (@arne) ?

cpp/include/generic_proxy_client.h
```diff
@@ -81,8 +82,14 @@ class GenericProxy_client : public Proxy_client {
   virtual void
   apply_sequence(const std::string &namespace_, const write_barrier,
                  const std::vector<std::shared_ptr<sequences::Assert>> &,
-                 const std::vector<std::shared_ptr<sequences::Update>> &,
-                 std::vector<proxy_protocol::object_info> &);
+                 const std::vector<std::shared_ptr<sequences::Update>> &);
+
+  // TODO want to make this a protected member but then rora_proxy_client can't use it
+  virtual void
+    apply_sequence_(const std::string &namespace_, const write_barrier,
```
domsj: added another make format

cpp/src/lib/osd_access.cc
```diff
@@ -31,19 +31,31 @@ OsdAccess &OsdAccess::getInstance() {
...
+  bool result = (map.find(osd) == map.end());
+  return result;
 }
 
-const std::map<osd_t, info_caps>::iterator OsdAccess::_find_osd(osd_t osd) {
-  std::lock_guard<std::mutex> lock(_osd_infos_mutex);
-  return _osd_infos.find(osd);
+std::shared_ptr<info_caps> OsdAccess::_find_osd(osd_t osd) {
+  std::lock_guard<std::mutex> lock(_alba_map_mutex);
+  auto &pair = _alba_map[_alba_map.size() - 1];
```
domsj: I changed the type at a certain point in time, I'll adapt some names too

cpp/src/lib/rora_proxy_client.cc
```diff
@@ -154,53 +166,154 @@ void RoraProxy_client::_maybe_update_osd_infos(
...
 }
 
-int RoraProxy_client::_short_path_one(const std::string & /*namespace_*/,
-                                      const ObjectSlices &object_slices,
-                                      std::shared_ptr<Manifest> mfp) {
+Location get_location(ManifestWithNamespaceId &mf, uint64_t pos, uint32_t len) {
+  int chunk_index = -1;
+  uint64_t total = 0;
+
+  {
```
domsj: it's not needed. the part in the block was in another function earlier on, so 'historic' reasons.
it releases the iterator a bit earlier, but I can remove it if you would prefer that.

cpp/src/lib/proxy_protocol.cc
```diff
@@ -391,32 +394,74 @@ void read_ping_response(message &m, Status &status, double &timestamp) {
...
-    std::vector<std::pair<osd_t, std::pair<std::unique_ptr<OsdInfo>,
-                                           std::unique_ptr<OsdCapabilities>>>> &
-        result) {
+void _read_osd_infos(message &m, osd_map_t &result) {
+  uint32_t n;
+  from(m, n);
+  for (uint32_t i = 0; i < n; i++) {
+    osd_t osd_id;
+    from(m, osd_id);
+    std::string info_s;
```
domsj: hadn't really given this any thought, it was already like this when I found it ;-), but I'll try to make an improvement here

cpp/src/lib/rora_proxy_client.cc
redlicha: As an exercise in code golf: :)

AFAICS `Location` does not have a move ctor / assigment operator, so there will be
* a copy into the temporary pair
* a copy for the intended move into the vector (btw, the constructed `pair` is already a "temporary" so there's no need for the `std::move`).
.
The latter could for example be avoided by using the `vector`'s `emplace` to construct the pair right into the vector - something along those lines:
```
results.emplace_back(target, get_location(...));
length -= results.back().length;
offset += results.back().length;
```
Taking a step back, the `results` vector is possibly resized, which will lead to more copies. This could be avoided by doing a `reserve` to the right size beforehand.

domsj:
Expanded the tests a bit, and fixed more errors. Everything should work now, except for the failures due to https://github.com/openvstorage/gobjfs/issues/12.

@redlicha can you make new volumedriver testers which integrate these changes? (api changed slightly, you'll see)

toolslive:
I think this can go in.

Issue #362, opened 2016-09-02
print proxy-client-cfg as json


jeroenmaelbrancke:
Can you add the option to get the proxy-client-cfg as json?

```
root@perf-roub-04:~# alba proxy-client-cfg --host 172.20.20.34 --port 26214
client_cfg:
{ cluster_id = "hdd_roub_abm";
  node_cfgs =
  [("03V0rdCWMNctuiu4", { ips = ["172.20.20.33"]; port = 26400 });
    ("Tqj1zMtIxTlijc5c", { ips = ["172.20.20.31"]; port = 26400 });
    ("ZVe2krRqMuP6ZQ1x", { ips = ["172.20.20.32"]; port = 26400 })];
  ssl_cfg = None;
  tcp_keepalive =
  { enable_tcp_keepalive = true; tcp_keepalive_time = 20;
    tcp_keepalive_intvl = 20; tcp_keepalive_probes = 3 }
  }
```

PR #398, opened 2016-10-12
Json of proxy cfg

fxfactorial:
Closes https://github.com/openvstorage/alba/issues/362

### DIFF
--- a/ocaml/.merlin
+++ b/ocaml/.merlin
@@ -1,8 +1,9 @@
 B _build
 B _build/src
+B _build/src/tools
 S src
 S src/kinetic
 PKG arakoon_client ctypes ocplib-endian 
 PKG lwt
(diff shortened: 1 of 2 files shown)


domsj:
Can you follow the style of the other commands a bit more closely?
Specifically, the other commands take a --to-json flag (see https://github.com/openvstorage/alba/blob/master/ocaml/src/cli_mgr.ml#L24 & https://github.com/openvstorage/alba/blob/master/ocaml/src/cli_mgr.ml#L52), which should then be used on line 474 in your case.

Doing it like this 1) keeps the output the same if --to-json argument is not given (you never know who might be crazy enough to parse it), and 2) in case of errors it will also be nicely wrapped in json output (try the command on a port where no proxy is running to see what I mean).

Issue #382, opened 2016-09-23
Many (1000) threads in asd process


domsj:
Most have the following stacktrace:

```
Thread 853 (Thread 0x7f1b6d7aa700 (LWP 18028)):
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x0000000000da5b65 in lwt_unix_condition_wait (condition=<optimized out>, mutex=<optimized out>) at src/unix/lwt_unix_stubs.c:263
#2  0x0000000000da5fbd in worker_loop (data=0x1e4fb90) at src/unix/lwt_unix_stubs.c:1214
#3  0x00007f1c1bf29184 in start_thread (arg=0x7f1b6d7aa700) at pthread_create.c:312
#4  0x00007f1c19fe537d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111

Thread 852 (Thread 0x7f1b6cfa9700 (LWP 18031)):
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x0000000000da5b65 in lwt_unix_condition_wait (condition=<optimized out>, mutex=<optimized out>) at src/unix/lwt_unix_stubs.c:263
#2  0x0000000000da5fbd in worker_loop (data=0x1e4fe00) at src/unix/lwt_unix_stubs.c:1214
#3  0x00007f1c1bf29184 in start_thread (arg=0x7f1b6cfa9700) at pthread_create.c:312
#4  0x00007f1c19fe537d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111

Thread 851 (Thread 0x7f1b6c7a8700 (LWP 18034)):
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x0000000000da5b65 in lwt_unix_condition_wait (condition=<optimized out>, mutex=<optimized out>) at src/unix/lwt_unix_stubs.c:263
#2  0x0000000000da5fbd in worker_loop (data=0x1e50070) at src/unix/lwt_unix_stubs.c:1214
#3  0x00007f1c1bf29184 in start_thread (arg=0x7f1b6c7a8700) at pthread_create.c:312
[...]

PR #402, opened 2016-10-17
configurable pool_size; default = 100

toolslive:
#382


domsj:
We haven't seen this with proxies/maintenance?
Makes me wonder why it would be triggered for the asds but not for those...
Anyway, looks good to me

toolslive:
no. only for ASDs.

PR #404, opened 2016-10-18
some exceptions shouldn't invalidate pool

toolslive:


ocaml/src/osd_access.ml
```diff
@@ -300,18 +300,19 @@ class osd_access
...
        | Exn Assert_failed _ -> false
        | _ -> true
      in
      let should_invalidate_pool =
        match exn with
-       (* TODO fill in exceptions for which we shouldn't invalidate the pool
-        *)
+       | Exn Assert_failed _ -> false
+       | Exn Unknown_operation -> false
+       | Exn Full -> false
```
domsj: While I agree that `Exn Full` shouldn't invalidate the pool, it should still somehow result in the osd not getting chosen for future writes, which as a result of this change is no longer the case.
So imo this line should be removed (or you'll need to change/add some logic elsewhere).

PR #405, opened 2016-10-18
we do not depend on cstruct

toolslive:


domsj:
typo ;-)
forgot the other docker files?

PR #409, opened 2016-10-18
Bench improvements

toolslive:
- `alba-bench` output of different clients to different files (and a process prefix)
   (we might want this for the other benchmarks too)
- `multi-delete` allows you to clean up after a killed osd bench

### DIFF
--- a/ocaml/src/cli_bench_common.ml
+++ b/ocaml/src/cli_bench_common.ml
@@ -45,11 +45,14 @@ let prefix default =
        & info ["prefix"] ~docv:"prefix" ~doc
   )
 
-let process_prefix default =
-  let doc = "$(docv) to keep multiple bench processes out of each other's way" in
+let client_file_prefix default =
+  let doc =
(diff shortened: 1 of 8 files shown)


ocaml/src/cli_bench_common.ml
```diff
@@ -45,6 +45,13 @@ let prefix default =
        & info ["prefix"] ~docv:"prefix" ~doc
   )
 
+let process_prefix default =
+  let doc = "$(docv) to keep multiple bench processes out of each other's way" in
```
domsj: If I understand the code correctly this argument is about keeping the log output of the different processes out of each other's way... At first I thought it was about the keys used on the asd (for which there is of course already another option).
Maybe make that a bit more clear in both the description and the variable name as it confused me?

toolslive: Indeed: every process will have one file per client, but if you have multiple processes, they will overwrite the files. Hence the prefix. I'll try to think of something.

PR #411, opened 2016-10-19
Extend get disk safety output

### LINKED ISSUE #403: Extend get-disk-safety json output
As discussed with @domsj.
Is it possible to extend the get-disk-safety on a backward compatible way with following information?

- add (updated) bucket lists including object count
- the amount of objects for the worst safety
- add a flag to simulate the disk safety with all the unavailable disks

domsj:
fixes #403


toolslive:
ok.

Issue #413, opened 2016-10-20
Claim osd looping and high cpu usage


JeffreyDevloo:
## Problem description
When I logged into my setup, I found that my root partition was completely full. The culprit was found in /var/log/upstart/alba-asd *. 
One of the asd logging was full of: 
```
2016-10-19 21:53:42 16108 +0200 - ovs-node1 - 9453/0 - alba/asd - 5366 - info - returning error (Asd_protocol.Protocol.Error.Assert_failed;    "p\000\000\000\000n\000\000\000\000g\000\000\000\000\000\000\000\000 \000\000\000\022v\0119\127\220C\168\166\216\005 \228\210\175\242\021\217\143aG\232A\158\184.\146\164\240g=a\142\159\003\000\002\000\000\000\000\000\000\000")
2016-10-19 21:53:42 17439 +0200 - ovs-node1 - 9453/0 - alba/asd - 5367 - warning - Assertion failed, expected some but got None instead
2016-10-19 21:53:42 17473 +0200 - ovs-node1 - 9453/0 - alba/asd - 5368 - info - returning error (Asd_protocol.Protocol.Error.Assert_failed;    "p\000\000\000\000n\000\000\000\000g\000\000\000\000\000\000\000\000 \000\000\000\022v\0119\127\220C\168\166\216\005 \228\210\175\242\021\217\143aG\232A\158\184.\146\164\240g=a\142\159\004\000\002\000\000\000\000\000\000\000")
2016-10-19 21:53:42 22216 +0200 - ovs-node1 - 9453/0 - alba/asd - 5369 - warning - Assertion failed, expected some but got None instead
2016-10-19 21:53:42 22241 +0200 - ovs-node1 - 9453/0 - alba/asd - 5370 - info - returning error (Asd_protocol.Protocol.Error.Assert_failed;
```
Every node within my cluster had an identical issue but the asd-id was different each time.

After closer inspection I spotted the following:
```
alba      9333 39.7  0.3 459100 52200 ?        Rsl  Oct19 380:40 /usr/bin/alba asd-start --config arakoon://config/ovs/alba/asds/0Z1uxwBY2djSHiAdG5xG855bwCHhZKfO/config?ini=%2Fopt%2Fasd-manager%2Fconfig%2Farakoon_cacc.ini --log-sink console:
```
The asd process is taking up alot of cpu.
The reason why is the claim-osd that is looping:
```
[...]

PR #416, opened 2016-10-21
block fragment_size<16 ; optimize chunk_size buffer for small objects

domsj:
fixes/prevents claim-osd looping seen in #413


ocaml/src/fragment_size_helper.ml
```diff
@@ -0,0 +1,50 @@
...
+*)
+
+(* this should be at least sizeof(long) according to the jerasure documentation *)
+let fragment_multiple = 16
+
+let round_down_to_multiple x y =
+  (x / y) * y
+
+let round_up_to_multiple x y =
+  if x = 0
```
toolslive:
```
let r = x mod y in
if r = 0 
then x 
else x + y - r  
``` ?

toolslive:
ok.

ocaml/src/fragment_size_helper.ml
domsj: ok, that's indeed a bit easier to follow

Issue #303, opened 2016-08-01
Allow more namespaces & work items


domsj:
Currently we internally use an int32 to track these, so after a while we may run out.
Running out isn't very likely, but in long enough living environments it could happen, and I would feel safer if we address this.
I think we can address this in a backwards compatible way, see https://github.com/openvstorage/alba/compare/extend-int32-to-int64 . That branch implements the needed changes for extending work ids, there's only a few TODOs left regarding testing (see the comments in prelude.ml).

See also https://github.com/openvstorage/openvstorage-health-check/issues/62#issuecomment-236190863. In the past we also had a bug which added work items at a rather high rate.

domsj:
fixed by #419

PR #419, opened 2016-10-24
Extend work item id & message id from int32 to int64

domsj:
(in a backwards compatible way)

this partially addresses #303 

equivalent change for namespace id is less urgent and will probably come later


ocaml/src/log_plugin.ml
```diff
@@ -28,41 +31,52 @@ let user_function (user_db : Registry.user_db) (value_o : string option) =
...
+  serialize
+    (Llio.tuple3_to
+       Llio.string_to
+       Llio.string_to
+       (Llio.list_to Llio.string_to))
+    (next_id_key, log_prefix, msgs)
+
+let make_update_x64 ~next_id_key ~log_prefix ~msgs =
+  let payload = make_payload ~next_id_key ~log_prefix ~msgs in
+  Update.Update.UserFunction (name_log_x64, Some payload)
```
domsj: This change of user function name forces that all nodes of the alba manager should be updated at the same time ...
So probably it's a better idea to keep the same name, which in newer version may write data in a different format (once id > Int32.max_int).

Hmm, now that I think of it: we should probably bump the albamgr version now too https://github.com/openvstorage/alba/blob/master/ocaml/src/albamgr_plugin.ml#L695 (as an old node may otherwise deserialize in an incorrect way leading to weird problems)?
(Again it would only happen when we actually ran out of work item ids)

So I guess that means they would want to update all abm nodes to the same version within some short timespan too.

toolslive: let's update albamgr version and never worry about this again.

toolslive:
ok, if jenkins likes it.

domsj:
Added some more commits so osd_id & namespace_id are now extended to an int64 too.
This PR now fully addresse #303.
Can you review those changes too?

domsj:
@redlicha can you make new voldrv testers containing these latest changes?

domsj:
new voldrv testers succeeded

toolslive:
ok.

PR #422, opened 2016-10-25
output default values in json representation of presets

domsj:
works around lack of https://github.com/whitequark/ppx_deriving_yojson/issues/19


toolslive:
ok (if compiles ;))

PR #430, opened 2016-10-28
add volume driver testers

toolslive:


domsj:
Started a run of alba_docker_generic_voldrv_backend on jenkins for ubuntu-16.04 (you only tried the alba_docker_generic_voldrv_tests job there), ok to merge if that one succeeds too

domsj:
The jenkins job succeeded, but this change breaks travis as it uses the ubuntu-16.04 dockerfile, which now needs access to the ovs lan.
Probably the easiest fix is using the centos-7 dockerfile for the tests running on travis?

toolslive:
what about testing on an env var that's only set by jenkins and not travis ?

domsj:
where would you test on an env var? inside the dockerfile?

toolslive:
There's something like `$docker build --build-arg <name>=<value> ...`
We can probably coerce Jenkins to do this. 
That would also allow everybody to use these docker files, not just us.

Issue #427, opened 2016-10-27
alba proxy-upload can fail after creating new namespace on global backend


JeffreyDevloo:
#### Problem description

The health check used to create only a single namespace for continuous use for every alba proxy. This is bad design as the health check should leave no traces of running once it has tested the required assets.

After implementing the removal of the namespace after the proxy test, we found out that the alba proxy-upload can fail for global backends (the error did not occur for local-backends).

Flow log that the health check has outputted:

```
[INFO] Checking ALBA proxy 'albaproxy_myvpool01': 
2016-10-26 10:38:12 07100 +0200 - ovs-node1 - 10988/139937514760000 - healthcheck/logger - 1 - INFO - Preset 'default' is not in use and will not be checked
[SKIPPED] Preset 'default' is not in use and will not be checked
2016-10-26 10:38:12 27300 +0200 - ovs-node1 - 10988/139937514760000 - healthcheck/logger - 2 - INFO - Namespace successfully created on proxy 'albaproxy_myvpool01' with preset 'mypreset'![SUCCESS] Namespace successfully created on proxy 'albaproxy_myvpool01' with preset 'mypreset'!
2016-10-26 10:38:12 54200 +0200 - ovs-node1 - 10988/139937514760000 - healthcheck/logger - 3 - ERROR - Failed to put object on namespace 'ovs-healthcheck-ns-mypreset' failed on proxy 'albaproxy_myvpool01' with preset 'mypreset' With error ''
[FAILED] Failed to put object on namespace 'ovs-healthcheck-ns-mypreset' failed on proxy 'albaproxy_myvpool01' with preset 'mypreset' With error ''
2016-10-26 10:38:12 54200 +0200 - ovs-node1 - 10988/139937514760000 - healthcheck/logger - 4 -INFO - Deleting namespace 'ovs-healthcheck-ns-mypreset'.
[INFO] Deleting namespace 'ovs-healthcheck-ns-mypreset'
```

A full debug log can be found here: http://pastebin.com/X7isPTea (posted on pastebin due to size of the log)

A quick analysis points towards these errors:
- Alba_client_errors.Error.NoSatisfiablePolicy
- Alba_client_errors.Error.NamespaceDoesNotExist (logged after creation of a namespace)

domsj:
The NamespaceDoesNotExist error is there because you tried to delete the object after already having deleted the namespace, so this one is expected.

wimpers:
[...]

PR #431, opened 2016-10-28
some logging improvements

domsj:
result of investigating #427 (now waiting on Jeffrey to provide me with new logs)


toolslive:
ok.

PR #432, opened 2016-11-02
fix for ubuntu 16.04 dockerfile

domsj:
#430 used a jenkins url with 'lastSuccessfulBuild' in it, these have a short lifetime.
Now using packages from a pinned build.


toolslive:
ok.

PR #436, opened 2016-11-03
Verify all dockerfiles on travis

domsj:
and fix the missing dependency it exposed for the centos-7 image

(also removed the travis caching effort as it doesn't work ... travis reports all cache as 0MB - so this was actually slowing down the builds)


toolslive:
ok.

PR #438, opened 2016-11-04
update to latest voldrv testers - now for centos-7 too

domsj:
(and the ubuntu-14.04 one no longer segfaults due to large namespace ids)


toolslive:
ok.

Issue #302, opened 2016-07-29
Lazy write out


wimpers:
We should optimize the write process by allowing a lazy write out where we acknowledge a write in case c fragments have been stored correctly. Currently we only do this in case of a disaster scenario but we should do this by default.

toolslive:
The primary use case in mind is a slow link to one of the data centers.

domsj:
Yes, and the secondary use cases would be 
- slow (deteriorating) disk (but just not slow enough)
- first failure of some osd

All can be nicely handled with the strategy discussed on the chat.

Actually this already helps for https://github.com/openvstorage/alba/issues/418 too (node goes down, takes some time to disqualify all osds) in case the disqualified node would still result in the same policy being chosen.

wimpers:
Maybe you can share that strategy with the rest of the world also?

toolslive:
Basically it's this:
- wait until `min_fragment_count fragments` have been written (this takes `t`)
- wait for a small arbitrary period (fe `0.2t`) for the slowpokes to arrive
- build the manifest with what you have at that point for the manifest and continue 
- if there are still outstanding writes that come in later, use those and update the manifest.

domsj:
We could also make the 0.2t configurable if desired.
Then you could e.g. have 0t in which case the upload has the lowest possible latency, or it could be 10000t in which case it would (probably) wait until the osd_timeout happens, or it could be some value chosen based on some testing for the specific environment.
(Making a good default choice for an arbitrary timeout is always hard...)

PR #439, opened 2016-11-07
Lazy write out

toolslive:
feature #302


ocaml/src/alba_bench.ml
```diff
@@ -75,6 +75,7 @@ let do_scenarios
   Alba_client2.with_client
     ~tcp_keepalive:Tcp_keepalive2.default
     ~populate_osds_info_cache:true
+    ~upload_slack:0.2
```
toolslive: yes. will make it configurable for alba bench too

ocaml/src/maintenance_test.ml
```diff
@@ -660,23 +760,27 @@ let test_verify_namespace () =
...
-             |> List.hd_exn
-             |> snd
-             = 1);
+     let new_version_missing = mf2.fragment_locations
+                               |> List.hd_exn
+                               |> List.hd_exn
+                               |> snd
+     in
+     OUnit.assert_bool "Manifest.version_id of formerly missing"
+       (1 <= new_version_missing);
```
domsj: should be exactly equal to 1? same remark for the other comparison below

ocaml/src/alba_test.ml
```diff
@@ -37,7 +37,16 @@ let test_with_alba_client ?bad_fragment_callback f =
         ~release_resources:true
         ~tcp_keepalive:Tcp_keepalive2.default
         ~populate_osds_info_cache:true
-        f
+        ~upload_slack:0.001
```
toolslive: yes. yes.

ocaml/src/tools/lwt_extra2.ml
```diff
@@ -298,3 +298,86 @@ let copy_between_fds fd_in fd_out size buffer =
   copy_using reader writer size buffer
 
 
+type 'a state =
```
toolslive: I'll have a look to simplify this.

ocaml/src/alba_client_upload.ml
```diff
@@ -419,18 +442,63 @@ let upload_object''
...
-           ~k ~m ~w'
+           ~k ~m ~w' ~min_fragment_count
            ~compression ~encryption ~fragment_checksum_algo
            ~version_id ~gc_epoch
            ~object_info_o
            ~osds:target_osds
            ~fragment_cache
-           ~cache_on_write)
+           ~cache_on_write
+           ~upload_slack
```
domsj: only add slack for the last chunk?
`~upload_slack:(if has_more then 0.0 else upload_slack`

toolslive: will do

ocaml/src/alba_osd.ml
```diff
@@ -136,7 +136,8 @@ class client
                    ~object_id_hint:None
                    ~fragment_cache:(alba_client # get_base_client # get_fragment_cache)
                    ~cache_on_write:(alba_client # get_base_client # get_cache_on_read_write |> snd)
-                 >>= fun (mf, _chunk_fidmos, _, gc_epoch) ->
+                   ~upload_slack:0.2
```
domsj: upload_slack in this class should depend on what's configured in proxy config file too?

ocaml/src/nsm_client.ml
```diff
@@ -150,9 +150,12 @@ class client (nsm_host_client : Nsm_host_client.basic_client) namespace_id =
...
         (allow_overwrite, object_name)
 
-    method update_manifest ~object_name ~object_id updated_object_locations ~gc_epoch ~version_id =
+    method update_manifest ~object_name ~object_id
+                           updated_object_locations
+                           ~gc_epoch ~version_id
+      =
       self # update
-        UpdateObject
+        UpdateObject2
```
toolslive: I'll adapt, but don't agree. Suppressing a call that no component is using is only hypothetically breaking backward compatibility.

ocaml/src/nsm_model.ml
```diff
@@ -1836,6 +1904,31 @@ module NamespaceManager(C : Constants)(KV : Read_key_value_store) = struct
...
+        kv object_name object_id
+        (new_fragments : (chunk_id
+                          * fragment_id
+                          * osd_id option
+                          * (int * Checksum.t) option)
+                           list)
+        gc_epoch
+        version_id
+    =
+    failwith "not implemented yet"
```
toolslive: it did fail.

ocaml/src/alba_osd.ml
toolslive: should it be the same as what's in the proxy's config ?

domsj:
There are some changes needed for recover_nsm_host.ml too: it previously assumed fragment_sizes & fragment_checksums in the recovery info would be filled up with the actual correct values.
As of this PR it can no longer assume this to be the case.
I think there are 2 options for fixing this:
- get the fragment size/checksum from the asd (query metadata about some key - we currently don't have this sort of operation but we can add it, for backwards compatibility we can get the same by downloading the entire fragment)
- we could enforce that the fragment checksum/size in each recovery info object is correct for the fragment the recovery info is associated with, the recovery code can then use this...

(my preference goes towards the second option)

PR #440, opened 2016-11-08
Proxy osd

domsj:


ocaml/src/nsm_model.ml
```diff
@@ -32,6 +32,23 @@ module OsdInfo = struct
...
+  type endpoint_uri = string [@@deriving show, yojson]
+
+  type host = string
+  type endpoint = Net_fd.transport * host * port
+
+  let parse_endpoint_uri s =
+    let uri = Uri.of_string s in
+    let transport = match Uri.scheme uri with
+      | None -> failwith "Transport should be 'tcp' or 'rdma'"
+      | Some "tcp" -> Net_fd.TCP
```
toolslive: what with upper case schemes ? Does the library convert to a canonical lower case string?

ocaml/src/proxy_osd.ml
```diff
@@ -21,18 +21,243 @@ open Slice
...
+      Lwt_list.map_p
+        (fun (transport, host, port) ->
+          Lwt.catch
+            (fun () ->
+              Lwt_unix.getaddrinfo host "" [ Unix.AI_SOCKTYPE Unix.SOCK_STREAM; ] >>= fun r ->
+              List.map
+                (fun addr_info ->
+                  transport,
+                  (match addr_info.Unix.ai_addr with
+                   | Unix.ADDR_UNIX x -> x
```
toolslive: will this work?

ocaml/src/proxy_osd.ml
```diff
@@ -21,18 +21,243 @@ open Slice
...
+
+let ensure_namespace_exists proxy_client ~namespace ~preset =
+  proxy_client # list_namespaces
+               ~first:namespace ~finc:true
+               ~last:(Some (namespace, true))
+               ~max:1 ~reverse:false
+  >>= fun ((_, x), _) ->
+  match x with
+  | [] -> proxy_client # create_namespace ~namespace ~preset_name:(Some preset)
+  | [ _; ] -> Lwt.return ()
```
toolslive: do we care that the preset of an existing namespace might be different?

ocaml/src/proxy_osd.ml
```diff
@@ -21,18 +21,243 @@ open Slice
...
+      Hashtbl.clear active_pools;
+      Hashtbl.iter (fun _ (_, fuse) -> fuse := true) disqualified_endpoints_requalify_threads;
+      Lwt.return ()
+  end
+
+let ensure_namespace_exists proxy_client ~namespace ~preset =
+  proxy_client # list_namespaces
+               ~first:namespace ~finc:true
+               ~last:(Some (namespace, true))
+               ~max:1 ~reverse:false
```
toolslive: convenience method on proxy_client to retrieve info of one namespace ?

ocaml/src/nsm_model.ml
domsj: I don't know of any automatic conversion that it does ... but the help message mentions 'tcp' and 'rdma' with lower case, so I don't see a problem.
Or are you trying to say it should support upper case schemes? (I don't see why we should.)

ocaml/src/proxy_osd.ml
domsj: ok

ocaml/src/nsm_model.ml
toolslive: Iirc, scheme names were supposed to be case insensitive. Hence the question.

ocaml/src/proxy_osd.ml
domsj: That would be rather unexpected ...
What would the scenario be where we get into such a situation?
- they deliberately already pre-create this namespace with some other preset
- they call add-osd with preset A, which succeeds in creating the namespace, but fails before adding the osd to the albamgr. Next they call add-osd again with another preset

One is sabotage, the other very unlikely. If you want I can add detection for it and fail with some error, in practice however I don't expect this to be a problem?

toolslive: I think it's better to fail if this happens.

ocaml/src/nsm_model.ml
domsj: Changed 'tcp' to 'tCp' in setup.ml and the test still seems to work, so it looks the library we use indeed does this conversion...
Fwiw see http://stackoverflow.com/questions/2148603/is-the-protocol-name-in-urls-case-sensitive

toolslive: ok.  from [RFC3986](https://tools.ietf.org/html/rfc3986:)
```
Although schemes are case-insensitive, the canonical form is lowercase and 
documents that specify schemes must do so with lowercase letters.  
An implementation should accept uppercase letters as equivalent to lowercase 
in scheme names (e.g., allow "HTTP" as well as "http") for the sake of
robustness but should only produce lowercase scheme names for consistency.
```
The library does the right thing ;)

ocaml/src/proxy_osd.ml
domsj: looks like this path shouldn't be triggered, I'll change it to `assert false`

toolslive:
ok.

PR #444, opened 2016-11-10
CTR mode encryption

domsj:
- better fit for partial reads
- no padding needed for encryption

### DIFF
--- a/ocaml/src/fragment_helper.ml
+++ b/ocaml/src/fragment_helper.ml
@@ -85,7 +85,9 @@ let maybe_encrypt
     get_iv algo key ~object_id ~chunk_id ~fragment_id ~ignore_fragment_id >>= fun iv ->
     Cipher.with_t_lwt
       key Cipher.AES256 Cipher.CBC []
-      (fun cipher -> Cipher.encrypt ~iv cipher bs) >>= fun () ->
+      (fun cipher ->
+        Cipher.set_iv cipher iv;
+        Cipher.encrypt cipher bs) >>= fun () ->
     Lwt.return bs
 
 let maybe_decrypt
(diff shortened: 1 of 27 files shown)


toolslive:
ok.

ocaml/src/fragment_helper.ml
```diff
@@ -104,13 +119,46 @@ let maybe_decrypt
...
+        get_iv algo key ~object_id ~chunk_id ~fragment_id ~ignore_fragment_id >>= fun iv ->
+        Cipher.with_t_lwt
+          key Cipher.AES256 Cipher.CTR []
+          (fun cipher ->
+            Cipher.set_ctr cipher iv;
+            Cipher.decrypt_detached
+                  cipher
+                  data 0 (Lwt_bytes.length data)
+          ) >>= fun () ->
+        Lwt.return (Bigstring_slice.wrap_bigstring data)
```
domsj: Fwiw it's not only the last step (unpadding) that differs, also the chosen mode (CBC/CTR) and set_iv/set_ctr differ. Lifting the commonalities highlights the differences, so I'll do that

domsj:
In https://github.com/openvstorage/alba/pull/439 it was discovered that the previous assumption that compression will always lead to the same output was wrong (compressors evolve).
In combination with deterministic IV/CTR generation and the CTR mode encryption implemented here this could (at least theoretically) leak information about the compressed data.
After some discussion with @toolslive about this on the chat we decided to generate a random IV/CTR upon encryption and store it in the manifest (and recovery info).

domsj:
@toolslive I added some more changes, should be finished now. Can you review again?

toolslive:
ok.

PR #448, opened 2016-11-15
catch ensure namespace race

domsj:
Avoids clusters of this error:
```
Nov 15 01:56:00 ftcmp02 alba[28561]: 2016-11-15 01:56:00 296291 +0100 - ftcmp02 - 28561/0 - alba/proxy - 114259 - info - Exception while adding object to alba fragment cache: Albamgr_protocol.Protocol.Error.Albamgr_exn(4, "Albamgr_protocol.Protocol.Error.Namespace_already_exists"); backtrace:; Raised at file "queue.ml", line 68, characters 17-22; Called from file "src/tools/lwt_pool2.ml", line 98, characters 25-46
```

### DIFF
--- a/ocaml/src/fragment_cache_alba.ml
+++ b/ocaml/src/fragment_cache_alba.ml
@@ -172,8 +172,14 @@ class alba_cache
        Lwt.catch
          (fun () -> f namespace)
          (fun exn ->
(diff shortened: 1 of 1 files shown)


toolslive:
ok.

PR #453, opened 2016-11-16
build/ci improvements

domsj:
- travis should fail (for all jobs) when the build fails
- don't build twice in packaging jobs
- cut down build time by combining build of alba.native & disk_failure_tests.native


toolslive:
ok.

PR #460, opened 2016-11-24
sleep in requalify-endpoint loop of proxy_osd

domsj:
Had to answer some questions about proxy_osd behavior for kenneth/wim/stefaan and found this issue while verifying afterwards if I hadn't told any lies.


toolslive:
ok.

PR #466, opened 2016-11-29
return codes for cli

toolslive:


setup/setup.ml
```diff
@@ -2009,6 +2021,35 @@ module Test = struct
...
 
+  let return_codes_1 t =
+    let cmd = [
+        "proxy-create-namespace";
+        "proxy_create_namespace_return_code"
+      ]
+    in
+    let out,rc = t.Deployment.proxy # cmd_line_with_capture_and_rc cmd in
+    let () = Printf.printf "output=%S\n%!" out in
+    assert (rc = 0);
```
toolslive: The first call creates a namespace and should succeed, hence `assert(rc = 0)`.
The second call tries to create the **same** namespace. This should fail and return with 
`rc = 2`. (`rc = 1` was reserved for input argument errors).

Issue #441, opened 2016-11-09
get-disk-safety --include-errored-as-dead returns weird results


JeffreyDevloo:
### Problem description
We were executing the following sequence to test get-disk-safety:
- Execute get-disk-safety
- Remove disks
- Execute get-disk-safety

After the removal of the disks we found:
```
root@ovs-node-1:~# alba get-disk-safety --config arakoon://config/ovs/arakoon/mybackend01-abm/config?ini=%2Fopt%2FOpenvStorage%2Fconfig%2Farakoon_cacc.ini --to-json --include-errored-as-dead
{
  "success": true,
  "result": [
    {
      "namespace": "0749d85b-4507-4093-a89f-3aebe3dfcd5f",
      "safety": 1,
      "safety_count": 1507,
      "bucket_safety": [
        {
          "bucket": [ 5, 4, 7, 3 ],
          "count": 14,
          "applicable_dead_osds": 3,
          "remaining_safety": -1
        },
        {
          "bucket": [ 5, 4, 8, 3 ],
          "count": 225,
          "applicable_dead_osds": 3,
          "remaining_safety": 0
        },
        {
          "bucket": [ 5, 4, 9, 3 ],
          "count": 1507,
          "applicable_dead_osds": 3,
          "remaining_safety": 1
        }
      ]
    },
    {
      "namespace": "4cda730a-1221-4047-84b4-c683c2b9b42f",
      "safety": 1,
      "safety_count": 31,
      "bucket_safety": [
        {
          "bucket": [ 5, 4, 9, 3 ],
          "count": 31,
          "applicable_dead_osds": 3,
          "remaining_safety": 1
        }
      ]
    },
    {
      "namespace": "fd-myvpool01-658e90b5-f392-457a-ad37-a1d142eb30d6",
      "safety": null,
      "safety_count": null,
      "bucket_safety": []
    }
  ]
}
```
and even later it got worse to -5 (but a get-disk-safety in between showed no errors):
```
root@ovs-node-1:~# alba get-disk-safety --config arakoon://config/ovs/arakoon/mybackend01-abm/config?ini=%2Fopt%2FOpenvStorage%2Fconfig%2Farakoon_cacc.ini --to-json --include-errored-as-dead
{
  "success": true,
  "result": [
    {
      "namespace": "0749d85b-4507-4093-a89f-3aebe3dfcd5f",
      "safety": 1,
      "safety_count": 1561,
      "bucket_safety": [
        {
[...]

PR #471, opened 2016-11-30
fix List.minima

domsj:
noticed something was wrong in https://github.com/openvstorage/alba/issues/441#issuecomment-263835263 and traced it back to this...


toolslive:
ok.

PR #481, opened 2016-12-06
test assert asd small value + fix bug it exposes

### LINKED ISSUE #478: osd message delivery too far off
saw this on @dejonghb's env. some osds kept on being disqualified due to this, and message delivery towards them got stuck

domsj:
really fixes #478


toolslive:
nice one!

Issue #446, opened 2016-11-10
Max fragments per node constraint violated


domsj:
@jeroenmaelbrancke claims to have seen the following on some env:
(5,4,8,3) policy, but somehow some objects were in a (5,4,9,4) bucket.
It's still unclear how this happened.
Some time later the maintenance process fixed this.
This could've been prevented by the check discussed in https://github.com/openvstorage/alba/issues/307.

domsj:
fixed by #486

Issue #307, opened 2016-08-05
changed policies are not propagated to the nsms


domsj:
Because the nsm doesn't know this new policy it blocks writes that may violate the policy (as a defency mechanism).

This resulted in the maintenance process not being able to do its work:

```
2016-08-05 09:34:45 6548 +0200 - perf-roub-04 - 20590/0 - alba/maintenance - 614580 - info - Exn while repairing object (~namespace_id:14178 ~object ~name:"r \000\000\000Z\142qU\004q\161\238\158h\154Iy\205
g{\152\224\209\1596\208\136\220w4\004G\159Mv\001\001\000\000\000\001\000\000\000\000\000\000\000" ~object_id:"\001y\240\254H\193L\208*\151\215\nPg'\150C3\022\173\018\144G\211\154\222\223r\213\216\219\018")
, will now try object rewrite: Nsm_model.Err.Nsm_exn(14, "Attempting to use 4 disks of node 5Yo6bZqwcZywSQZlXE3aU6pZLfAWLTC5 while max 3 permitted ([(Some 17l); (Some 7l); (Some 25l); (Some 14l); (Some 1l)
; (Some 24l);\n (Some 12l); (Some 20l); (Some 2l); (Some 18l); (Some 27l); (Some 4l)])"); backtrace:; Raised at file "queue.ml", line 86, characters 29-34; Called from file "src/tools/lwt_pool2.ml", line 9
8, characters 25-46
```

It could probably also result in writes from the proxy failing due to this.

Workaround: don't change your policy, and if you did change it back to the original value.

domsj:
The issue is actually even worse: once an object has been written with a certain amount of disks per node it can never get a higher amount of disks per node, regardless of what is in the policy.
The policies ain't even propagated to the nsm when it's created...

domsj:
reducing priority to normal because #309 has been merged

wimpers:
After discussion with @domsj  and @toolslive  moved to RC2

domsj:
fixed by #486

PR #486, opened 2016-12-07
Nsm validate preset

domsj:
Fixes the following issues:
https://github.com/openvstorage/alba/issues/307
https://github.com/openvstorage/alba/issues/446


ocaml/src/albamgr_client.ml
```diff
@@ -229,6 +229,23 @@ object(self)
         ~max:1 ~reverse:false >>= fun ((_, presets), _) ->
       Lwt.return (List.hd presets)
 
+    method list_presets2 ~first ~finc ~last ~reverse ~max =
+      client # query
+        ListPresets2
```
toolslive: what happens if the albamgr does not support this call?

domsj: This call should only be used in code paths related to this new functionality.
So only the new functionality shouldn't work.
However I now see that anywhere a preset gets refreshed will stop working...
I'll change the return type here so that instead of `'a` it returns `'a option` and deal with that at the call sites so we can get the desired behaviour.

ocaml/src/albamgr_client.ml
```diff
@@ -229,6 +236,29 @@ object(self)
...
+    method get_preset2 ~preset_name =
+      self # list_presets2
+           ~first:preset_name ~finc:true ~last:(Some (preset_name, true))
+           ~max:1 ~reverse:false >>= function
+      | None ->
+         Lwt.return_none
+      | Some ((_, presets), _) ->
+         Lwt.return (List.hd presets)
+
+    method get_preset_propagation_state ~preset_name =
```
toolslive: Is there value in a command line call for this?

toolslive:
ok.

toolslive:
ok.

PR #488, opened 2016-12-07
Fix #487 proxy apply_sequence doesn't update manifest cache

### LINKED ISSUE #487: proxy apply_sequence doesn't update manifest cache
So after an overwrite of an object the old manifest may be used for inconsistent reads, resulting in wrong data being returned to the client.

Actually all code in Alba_client_upload.store_manifest is skipped in this code path.
This means that besides not updating the manifest cache we're also not cleaning up gc tags & not doing the lazy writeout epilogue.
(Luckily those last 2 aren't needed for correctness, but still we should fix this.)

domsj:
Also, it didn't clear the manifest cache when deleting an object, and this issue was present in the alba_osd code too. So extracted the commonalities and fixed that too.

### DIFF
--- a/ocaml/src/alba_client_sequence.ml
+++ b/ocaml/src/alba_client_sequence.ml
@@ -1,0 +1,113 @@
+(*
+Copyright (C) 2016 iNuron NV
+
+This file is part of Open vStorage Open Source Edition (OSE), as available from
+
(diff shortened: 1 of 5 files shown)


toolslive:
ok.

PR #501, opened 2016-12-14
Improvements related to #427

domsj:
See #427.
The added test could first reproduce this, and after the changes to message delivery on namespace creation it now succeeds (most of the time, see message about segfault below).
Of course @JeffreyDevloo will want to verify this too: you can find a fresh package @ http://10.100.129.100:8080/view/alba2/job/alba_docker_generic_package_ubuntu-16.04/lastSuccessfulBuild/artifact/alba_1.2.4-83-g9da540f_amd64.deb

In an unexpected turn of events the new test also allows reproducing a segfault which I still have to investigate - but which I believe to be unrelated to this pull request.


toolslive:
ok.

PR #503, opened 2016-12-15
cleanup of fabric code

domsj:


toolslive:
ok, if integration tests succeed.

PR #504, opened 2016-12-15
recycle buffers in asd_client

toolslive:


ocaml/src/tools/llio2.ml
```diff
@@ -21,22 +21,26 @@ open Lwt_bytes2
...
+               mutable pos : int;
+               too_far : int;
+             }
 
     type 'a deserializer = t -> 'a
 
-    let make_buffer buf pos = { buf; pos; }
+    let make_buffer buf pos too_far = { buf; pos; too_far }
 
     let advance_pos buf delta = buf.pos <- buf.pos + delta
```
domsj: add an assert here that `buf.pos <= buf.max_pos`
(max_pos is more descriptive for me than too_far)

toolslive: ok

ocaml/src/tools/llio2.ml
```diff
@@ -21,22 +21,28 @@ open Lwt_bytes2
...
+             }
 
     type 'a deserializer = t -> 'a
 
-    let make_buffer buf pos = { buf; pos; }
+    let make_buffer buf pos max_pos =
+      assert (pos <= max_pos);
+      { buf; pos; max_pos }
 
     let advance_pos buf delta = buf.pos <- buf.pos + delta
```
domsj: you still want to add an assert here too

domsj:
it probably makes sense to release (unsafe_destroy) the buffer when the client is disposed, otherwise we'll introduce a space leak

domsj:
ok

PR #506, opened 2016-12-15
be a bit more persistent about fragment cache namespace cleanup

domsj:
Some more description in the code comment. Saw this problem on the env of @dejonghb.


toolslive:
ok

PR #511, opened 2016-12-19
Ctr c++

toolslive:


cpp/src/lib/proxy_protocol.cc
```diff
@@ -289,18 +289,29 @@ void _read_object_infos(message &m, std::vector<object_info> &object_infos) {
...
-    auto t =
-        std::make_tuple(std::move(name), std::move(future), std::move(umf));
-    object_infos[i] = std::move(t);
+    try {
+      unique_ptr<ManifestWithNamespaceId> umf(new ManifestWithNamespaceId());
+      from(m, *umf);
+      assert(name == umf->name);
+      auto t = make_tuple(move(name), move(future), move(umf));
+      object_infos.push_back(move(t));
+    } catch (alba::llio::deserialisation_exception &e) {
```
domsj: https://github.com/openvstorage/alba/blob/bcacbf70ba0a47378883a2681df243b9148031d9/cpp/src/lib/manifest.cc#L147
that line also throws this kind of exception, and at that point we can't yet skip the manifest...
so we may want to make a distinction between those?
or should the proxy never return such manifests as the result of a ReadObjectSlices / ApplySequence call?
(in that last case we may want to leave a comment at some appropriate place(s)...)

toolslive: perhaps it's better to assert that the manifest version is 1.

toolslive: `ok_to_continue` it is.

Issue #461, opened 2016-11-25
Proxy client connections hanging


domsj:
This happened on @dejonghb's env 

Setup: accelerated backend

Around 10:04 the node hosting (half of the) asds of the fragment cache backend went down.

In the log of the proxy of the main backend we then see this:

```
Nov 25 10:05:41 ftcmp02 alba[22625]: 2016-11-25 10:05:41 368610 +0100 - ftcmp02 - 22625/0 - alba/proxy - 442472 - info - server: new client connection from (172.22.186.32,35092)
Nov 25 10:05:42 ftcmp02 alba[22625]: 2016-11-25 10:05:42 193100 +0100 - ftcmp02 - 22625/0 - alba/proxy - 442473 - info - server: new client connection from (172.22.186.32,35136)
Nov 25 10:05:45 ftcmp02 alba[22625]: 2016-11-25 10:05:45 72825 +0100 - ftcmp02 - 22625/0 - alba/proxy - 442474 - info - server: new client connection from (172.22.186.32,35590)
Nov 25 10:05:45 ftcmp02 alba[22625]: 2016-11-25 10:05:45 73584 +0100 - ftcmp02 - 22625/0 - alba/proxy - 442475 - info - server: new client connection from (172.22.186.32,35592)
Nov 25 10:05:57 ftcmp02 alba[22625]: 2016-11-25 10:05:57 369509 +0100 - ftcmp02 - 22625/0 - alba/proxy - 442476 - info - server: new client connection from (172.22.186.32,37264)
Nov 25 10:05:58 ftcmp02 alba[22625]: 2016-11-25 10:05:58 193949 +0100 - ftcmp02 - 22625/0 - alba/proxy - 442477 - info - server: new client connection from (172.22.186.32,37434)
Nov 25 10:06:01 ftcmp02 alba[22625]: 2016-11-25 10:06:01 73753 +0100 - ftcmp02 - 22625/0 - alba/proxy - 442478 - info - server: new client connection from (172.22.186.32,37904)
```

and a bit later it turns in to

```
Nov 25 10:15:50 ftcmp02 alba[22625]: 2016-11-25 10:15:50 104537 +0100 - ftcmp02 - 22625/0 - alba/proxy - 442600 - info - server: new client connection from (172.22.186.32,40850)
[...]

PR #515, opened 2016-12-19
use default tcp keepalive for client connections

domsj:
This may fix https://github.com/openvstorage/alba/issues/461.
There are some paths in (partial) download in which no timeouts are set, which could - in combination with the lack of tcp keepalive - explain #461.


toolslive:
ok.

PR #530, opened 2016-12-22
Osd view fix

toolslive:


ocaml/src/tools/llio2.ml
```diff
@@ -55,6 +55,10 @@ module ReadBuffer = struct
         let res = a_from buf in
         assert (buffer_done buf);
         res
+    let maybe_from_buffer_no_assert a_from default buf =
```
domsj: interesting fix, but don't we want to remove the assert everywhere?
it limits the extensibility of our serialisation ...
or maybe not remove it everywhere but at least evaluate all places that use maybe_from_buffer?

domsj: fwiw the equivalent assert in Prelude.maybe_from_buffer was already removed some time ago, see https://github.com/openvstorage/alba/commit/555d872dcf76355f6367c2f44ed73c6837d5999a

toolslive: I'll just kill the assert.

Issue #529, opened 2016-12-22
lazy writeout leads to unnecessary repair load


domsj:
While the last fragments for an object are still being written, but after the manifest was already stored in the nsm, an object may be picked up by maintenance (repair_by_policy/bucket) for repair.

This repair is (most of the time) not needed and thus a waste of resources.

repair_by_policy should skip objects with version=0 that have been written during e.g. the last minute.
(Do note that we don't want to get into a situation where a constant stream of new object writes prevents repair_by_policy from doing any work!)

I can probably pick this up in my https://github.com/openvstorage/alba/tree/remove-k-c-restriction branch.

domsj:
fixed by #532

Issue #516, opened 2016-12-20
Remove  k <= c restriction


wimpers:
Remove the restriction k <= c  for policies in an ALBA backend.

saelbrec:
In case maintenance would have to repair a missing data fragment after a proxy failed, a policy where c < k would lead to dataloss. This might be acceptable for an alba backend being used as fragment cache but for sure not advisable for any others.

What do we actually want to achieve with this 'enhancement' ?

wimpers:
After discussion with stakeholders: don't remove the restriction.

Issue #437, opened 2016-11-04
Investigate behaviour of maintenance process in case there are unavailable/broken objects


domsj:
We want to make sure that maintenance does not start spinning, it can retry once every minute or so though...
If the namespace is used as a cache: maybe delete the object?

wimpers:
Do you have a specific test case in mind? is this something QA needs to pickup or just a stub for yourself.

domsj:
shut down a few too many asds, see what maintenance does
purge a few osds (so many that the gui asks to confirm you're willing to accept dataloss), see what maintenance does

(need to wait at least 1 minute for the reaction from maintenance)

we can do this ourselves or QA could do it
(I was thinking of doing it myself one of these days)

not sure yet if I can/want to make this an automatic test

wimpers:
please investigate

dejonghb:
accidentally investigated this on my env.
had dataloss due to removal of too many disks, and saw 3500+ gets/s on the asds. maintenance was consuming 100%+ cpu.
after removing the namespaces which had dataloss (and restarting maintenance) the activity stopped.

domsj:
additionally we noticed that purging osds were not yet fully removed
(nsm_client # update_manifest failed with Insufficient_fragments ... because the nsm won't tolerate dataloss)

domsj:
fixed by #532

PR #532, opened 2016-12-26
Handle broken/unavailable objects a bit better

domsj:


domsj:
This addresses #437 & #529.
The change requested in #516 has not been made, because people don't know yet what they want

Issue #537, opened 2016-12-30
Failure "can't happen" exception


dejonghb:
Found 12 messages as this little gem in the logs:
```
Dec 29 16:50:24 ftcmp02 alba[11847]: 2016-12-29 16:50:24 142163 +0100 - ftcmp02 - 11847/0 - alba/proxy - 11068 - info - Exception while adding object to alba fragment cache: (Failure "can't happen"); backtrace:; Raised at file "format.ml", line 241, characters 41-52; Called from file "format.ml", line 482, characters 6-24
```
The proxy was running in debug mode and 7 of these are on debug level, 5 on info.

domsj:
On the same env we also saw https://github.com/openvstorage/alba/issues/209 .
This issue is probably caused by the other issue.
(The logic in choose.ml doesn't like negative fill rates...)

Issue #209, opened 2016-05-26
Negative usage on asds


jeroenmaelbrancke:
```
alba version
0.9.5
git_revision: "tags/0.9.5-0-ge4cf366"
git_repo: "https://github.com/openvstorage/alba.git"
compile_time: "11/05/2016 15:40:26 UTC"
machine: "455badec0dde 3.13.0-24-generic x86_64 x86_64 x86_64 GNU/Linux"
compiler_version: "4.02.3"
```

On regular time intervals, asds/backend usage shows negative values. (see screenshots)
After restarting the asd the negative value disappears and the correct value is shown.

Negative used_bytes:

```
root@str02:~# alba list-osds --config etcd://127.0.0.1:2379/ovs/arakoon/be-backend-abm/config | grep oyR
2016/05/26 15:20:10 286478 +0200 - str02 - 1403/0 - alba/cli - 0 - info - ETCD: etcdctl --peers=127.0.0.1:2379 get ovs/arakoon/be-backend-abm/config                             "oyRgZLZKDUVXts1birJ1IceDxUJ32EsH");
  other = "{ \"id\" : \"oyRgZLZKDUVXts1birJ1IceDxUJ32EsH\", \"node_id\" : \"qNqsb2zoMLDE7iCAkSBpNz24a27iwDNX\", \"port\" : 8603, \"used_bytes\" : \"-187152090328\", \"total_bytes\" : \"1999018582016\", \"version\" : \"AsdV1\", \"network_interfaces\":[ ]}";
```

Restarting asd (correct value):

```
root@str02:~# restart alba-asd-oyRgZLZKDUVXts1birJ1IceDxUJ32EsH
alba-asd-oyRgZLZKDUVXts1birJ1IceDxUJ32EsH start/running, process 10051

root@str02:~# alba list-osds --config etcd://127.0.0.1:2379/ovs/arakoon/be-backend-abm/config | grep oyR
2016/05/26 15:23:05 632690 +0200 - str02 - 11000/0 - alba/cli - 0 - info - ETCD: etcdctl --peers=127.0.0.1:2379 get ovs/arakoon/be-backend-abm/config                             "oyRgZLZKDUVXts1birJ1IceDxUJ32EsH");
[...]

PR #539, opened 2017-01-02
choose lives on bad input; asd dies on neg usage

toolslive:


toolslive:
awaiting jenkins result.

domsj:
should fix #537 & #209

PR #544, opened 2017-01-04
C++ asd client (rora alternative)

domsj:


cpp/src/lib/osd_access.cc
```diff
@@ -117,16 +117,45 @@ int OsdAccess::read_osds_slices(
     osd_t osd = item.first;
     auto &osd_slices = item.second;
     // TODO this could be done in parallel
-    rc = _read_osd_slices(osd, osd_slices);
+    rc = _read_osd_slices_asd_direct_path(osd, osd_slices);
```
domsj: this change means that when using the rora proxy client we'll currently always connect directly to the asd using the ocaml protocol and never via the gobjfs/libxio based protocol

cpp/src/lib/asd_access.cc
```diff
@@ -0,0 +1,165 @@
...
+
+void ConnectionPool::clear_(Connections &conns) {
+  while (not conns.empty()) {
+    pop_(conns);
+  }
+}
+
+std::unique_ptr<Asd_client> ConnectionPool::make_one_() const {
+  auto duration = std::chrono::seconds(5);
+  std::unique_ptr<transport::Transport> transport(new transport::TCP_transport(
```
domsj: this should look at the transport configured for this asd and possibly instantiate an RDMA_transport here

domsj:
fixes https://github.com/openvstorage/alba/issues/545

cpp/include/asd_access.h
```diff
@@ -0,0 +1,93 @@
...
+
+  ConnectionPools() {}
+
+  ConnectionPools(const ConnectionPools &) = delete;
+
+  ConnectionPools &operator=(const ConnectionPools &) = delete;
+
+private:
+  mutable alba::spinlock::SpinLock lock_;
+  std::map<std::string, ConnectionPool *> connection_pools_;
```
redlicha: `std::map<std::string, std::unique_ptr<ConnectionPool>>` would make the `delete` in the dtor / the whole dtor obsolete.

cpp/include/asd_client.h
```diff
@@ -0,0 +1,59 @@
...
+
+  virtual const char *what() const noexcept { return _what.c_str(); }
+};
+
+class Asd_client : public boost::intrusive::slist_base_hook<> {
+public:
+  Asd_client(const std::chrono::steady_clock::duration &,
+             std::unique_ptr<transport::Transport> &&);
+
+  void init(boost::optional<string> long_id);
```
redlicha: With the current usage, `init` should be folded into the ctor - cf. https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c41-a-constructor-should-create-a-fully-initialized-object

cpp/include/asd_protocol.h
```diff
@@ -0,0 +1,65 @@
...
+};
+
+struct slice {
+  uint32_t offset;
+  uint32_t length;
+  byte *target;
+};
+
+static string _MAGIC = "aLbA";
+static uint32_t _VERSION = 1;
```
redlicha: These create instances in each compilation unit including the header.

cpp/include/asd_access.h
```diff
@@ -0,0 +1,93 @@
...
+  static void clear_(Connections &);
+};
+
+class ConnectionPools {
+public:
+  ConnectionPool *get_connection_pool(const proxy_protocol::OsdInfo &);
+
+  ~ConnectionPools();
+
+  ConnectionPools() {}
```
redlicha: Could use `ConnectionPools() = default` instead.

cpp/include/spinlock.h
```diff
@@ -0,0 +1,69 @@
...
+  explicit ScopedSpinLock(SpinLock &sl);
+
+  ScopedSpinLock(const ScopedSpinLock &) = delete;
+  ScopedSpinLock &operator=(const ScopedSpinLock &) = delete;
+
+  ~ScopedSpinLock();
+
+private:
+  SpinLock &sl_;
+};
```
redlicha: I'd suggest to not even start getting `ScopedSpinLock` over into Alba but to stick to `std::lock_guard` and friends right away.

domsj:
@redlicha took your comments into account, feel free to have another look

Will now have a look at implementing the circuit breaker

domsj:
notes to self:
- compile with clang too and address warnings
- license headers

PR #553, opened 2017-01-06
proxy shouldn't try to discover asds

domsj:
the maintenance will still do it, and we don't depend on it for correct functioning of the proxy


domsj:
hold on, proxy_osd_tests are failing ... will to investigate

PR #555, opened 2017-01-06
Helgrind

toolslive:


toolslive:
```
==28769== Helgrind, a thread error detector
==28769== Copyright (C) 2007-2015, and GNU GPL'd, by OpenWorks LLP et al.
==28769== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==28769== Command: ./cpp/bin/test_client.out --host 192.168.0.212 --port 10000 --command partial-read-benchmark --benchmark-size 10000 --namespace rora_bench --file /home/romain/workspace/ALBA/alba_bb/32MB.bin --n-clients 10 --use-rora=true --io-pattern=random --invalidate-cache=true --log-level=info
==28769== 
==28769== 
==28769== For counts of detected and suppressed errors, rerun with: -v
==28769== Use --history-level=approx or =none to gain increased speed, at
==28769== the cost of reduced accuracy of conflicting-access information
==28769== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 6193725 from 230)
```

cpp/src/lib/asd_access.cc
```diff
@@ -19,14 +19,13 @@
...
 
-#include <boost/thread/lock_guard.hpp>
+#include <mutex>
 
 namespace alba {
 namespace asd {
 
 using alba::proxy_protocol::OsdInfo;
 
-#define LOCK() boost::lock_guard<decltype(lock_)> lg__(lock_)
```
domsj: I liked this LOCK macro

cpp/src/lib/manifest_cache.cc
```diff
@@ -30,6 +30,7 @@ ManifestCache &ManifestCache::getInstance() {
 
 static size_t _manifest_cache_capacity = 10000;
```
domsj: this should be a class member then iso static?

toolslive: doesn't really matter as long as the access to it is guarded.

domsj:
right, don't forget to remove spinlock from https://github.com/openvstorage/alba/blob/helgrind/cpp/automake_client/Makefile.am

PR #563, opened 2017-01-11
Setup cpp global nil

toolslive:


setup/setup.ml
```diff
@@ -1689,48 +1690,181 @@ module Test = struct
     Deployment.setup ~bump_ids t_hdd;
     (cfg_hdd, t_hdd), (cfg_ssd, t_ssd)
 
-  let cpp ?(xml=false) ?filter ?dump (_:Deployment.t) =
-    let (_, t_hdd), (_, t_ssd) = setup_aaa ~bump_ids:true ~the_preset:"preset_rora" () in
 
-    t_hdd.Deployment.osds.(0) # set "key1" (String.make 100 'a');
+  let setup_global_backend () =
```
domsj: There are now 2 definitions of setup_global_backend in this file, You probably forgot to remove the one below? (so this would result in a move of the function iso a copy)

toolslive: a merge accident?

toolslive:
I'll add the other cases too.

PR #570, opened 2017-01-11
less lwt cancel & friends

domsj:


domsj:
Hmm, travis failed due to too many open fds. Could very well be because of the changes here, so we might want to hold off with merging.
(I'll investigate once I'm back)

PR #572, opened 2017-01-13
recycle buffer

toolslive:


cpp/include/llio.h
```diff
@@ -126,25 +128,22 @@ class message {
   size_t _size;
 };
 
+#define _SIZE0 32
```
redlicha: a `static constexpr size_t size0 = 32` in `message_builder` (private?) is preferrable over `#define`'s

cpp/include/llio.h
```diff
@@ -157,19 +156,44 @@ class message_builder {
     }
   }
 
-  void add_raw(const char *b, uint32_t size) noexcept {
-    _buffer.write(b, size);
+  void add_raw(const char *b, uint32_t len) noexcept {
+      uint free = _size - _pos;
+      if(free < len){
```
redlicha: If this resize path is not hit too frequently, a `std::vector<char>` should make things simpler: no need to track resources / take care of copying.

Issue #585, opened 2017-01-23
Inconsistent errors when claiming already claimed OSDs 


kvanhijf:
Through the GUI i claimed an ASD on some backend, which was successful, but when trying to claim the same ASD on another backend i get 2 different errormessages

1st error message:
```
root@OVS-1-193-151:~# /usr/bin/alba claim-osd  --config=arakoon://config/ovs/arakoon/bend2-abm/config?ini=%2Fopt%2FOpenvStorage%2Fconfig%2Farakoon_cacc.ini --long-id=Gr59Ibls3QClaCcHh6YgPAt85y3Ovt0O
This OSD is already claimed by alba instance 47b887a4-770c-47c4-8027-44090c3c9098
2017-01-23 09:41:50 731159 +0100 - OVS-1-193-151 - 6205/0 - alba/cli - 0 - warning - (Failure;   "Failed to add OSD because it's already claimed by another alba instance")
```

Subsequent error messages:
```
root@OVS-1-193-151:~# /usr/bin/alba claim-osd  --config=arakoon://config/ovs/arakoon/bend2-abm/config?ini=%2Fopt%2FOpenvStorage%2Fconfig%2Farakoon_cacc.ini --long-id=Gr59Ibls3QClaCcHh6YgPAt85y3Ovt0O
2017-01-23 09:50:19 251686 +0100 - OVS-1-193-151 - 12289/0 - alba/cli - 0 - warning - Albamgr exception: Albamgr_protocol.Protocol.Error.Osd_already_claimed
```

The framework expects an error in a dictionary format, to be JSON parsable, but we're getting a string instead (We always pass the --to-json flag)

toolslive:
the cli command you're using doesn't include `--to-json` (even though you say you pass it in)
if you do so, you get:
```
$>  alba claim-osd --to-json --long-id 8000_00_0ae84aaf --config ... 2> /dev/null
{"success":false,
 "error":{"message":"Albamgr exception(Albamgr_protocol.Protocol.Error.Osd_already_claimed,Albamgr_protocol.Protocol.Error.Osd_already_claimed)","exception_type":"albamgr_exn","exception_code":11}}
```
from which you can get the `exception code`

PR #586, opened 2017-01-23
use stderr iso stdout when printing message

toolslive:
fixes ( #585 )

### DIFF
--- a/ocaml/src/alba_client_osd.ml
+++ b/ocaml/src/alba_client_osd.ml
@@ -122,7 +122,7 @@ let claim_osd mgr_access osd_access ~long_id =
                  "This OSD is already claimed by alba instance %s"
                  alba_id'
      in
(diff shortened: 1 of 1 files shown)


ocaml/src/alba_client_osd.ml
```diff
@@ -118,8 +118,12 @@ let claim_osd mgr_access osd_access ~long_id =
      update_osd ())
   >>= function
   | `ClaimedBy alba_id' ->
-     Lwt_io.printlf "This OSD is already claimed by alba instance %s" alba_id' >>= fun () ->
-     Lwt.fail_with "Failed to add OSD because it's already claimed by another alba instance"
+     let msg = Printf.sprintf
+                 "This OSD is already claimed by alba instance %s"
+                 alba_id'
+     in
+     Lwt_io.eprintl msg >>= fun () ->
```
domsj: We should just use Lwt_log here.

PR #604, opened 2017-02-06
refactor c++ proxy client to use 'transport'

domsj:


toolslive:
ok, (but move to c++14 )

PR #605, opened 2017-02-06
cancel refreshers on finalization

toolslive:
This makes us leak way less (max < 600 fds) in test cases.

### DIFF
--- a/ocaml/src/osd_access.ml
+++ b/ocaml/src/osd_access.ml
@@ -255,6 +255,7 @@ class osd_access
     (Hashtbl.create 3
      : (Osd.id,
         OsdInfo.t * Osd_state.t  * Capabilities.OsdCapabilities.t) Hashtbl.t) in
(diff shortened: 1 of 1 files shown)


ocaml/src/osd_access.ml
```diff
@@ -313,7 +315,13 @@ class osd_access
         | None -> failwith (Printf.sprintf "could not find osd with id %Li" osd_id)
         | Some info -> info
       in
-      add_osd_info ~osd_id  osd_info Capabilities.OsdCapabilities.default;
+      let maybe_refresher =
+        add_osd_info ~osd_id  osd_info Capabilities.OsdCapabilities.default
+      in
+      let () = match maybe_refresher with
+      | None -> ()
+      | Some r -> r () |> Lwt.ignore_result
```
domsj: this path should add an item to the refreshers list too?

PR #614, opened 2017-02-10
timeouts fragment cache lookup; download fragment

toolslive:


ocaml/src/fragment_cache.ml
```diff
@@ -32,8 +32,11 @@ class virtual cache = object(self)
     method add' bid oid blob =
       self # add bid oid blob >>= fun _ -> Lwt.return_unit
 
-    method virtual lookup : int64 -> string -> (Lwt_bytes.t * (Manifest.t * int64 * string) list) option Lwt.t
-    method virtual lookup2 : int64 -> string
+    method virtual lookup : ?timeout:float
+                            -> int64 -> string
+                            -> (Lwt_bytes.t * (Manifest.t * int64 * string) list) option Lwt.t
+    method virtual lookup2 : ?timeout:float
```
domsj: Did you intend on using this extra argument? (Currently it's not used anywhere.)

ocaml/src/fragment_cache.ml
```diff
@@ -438,24 +441,33 @@ class blob_cache root ~(max_size:int64) ~rocksdb_max_open_files
...
-           bid oid
-           (fun fd ~len ->
-            Lwt_list.iter_s
-              (fun (offset, length, target, offset') ->
-               Lwt_unix.lseek fd offset Lwt_unix.SEEK_SET >>= fun _ ->
-               Lwt_extra2.read_all_lwt_bytes_exact fd target offset' length)
-              slices) >>= function
-      | Some () -> Lwt.return (true, [])
-      | None -> Lwt.return (false, [])
+    method lookup ?(timeout= 10.0) bid oid =
```
domsj: Does having a hard coded value make sense here? Maybe it's better to force the caller to supply a reasonable value.
Fwiw there's another hard coded value in fragment_cache_alba.ml.

toolslive: ok, will make it a non-optional label.

PR #621, opened 2017-02-17
Manifest serialization

toolslive:


ocaml/src/nsm_host_client.ml
```diff
@@ -167,17 +172,37 @@ class client (client : basic_client) =
...
       client # update (NsmsUpdate Nsm_protocol.Protocol.UpdatePreset) req
+
+    method update_session kvs =
+      client # query UpdateSession kvs
   end
 
 let wrap_around (client:Arakoon_client.client) =
-  client # user_hook "nsm_host" >>= fun (ic, oc) ->
+  let hook_name = "nsm_host" in
+  let session = Nsm_protocol.Session.make () in
```
domsj: Looks like the server side session isn't needed on the client side?

ocaml/src/nsm_host_plugin.ml
```diff
@@ -375,6 +378,24 @@ let handle_query : type i o. read_user_db -> (i, o) Nsm_host_protocol.Protocol.q
...
+     begin
+       List.iter
+         (function
+          | "manifest_ser", vo ->
+             begin
+               match vo with
+               | None -> ()
+               | Some v ->
+                  let new_version = deserialize Llio.int8_from v in
+                  Nsm_protocol.Session.set_manifest_ser session new_version
```
domsj: What are the exact semantics of the manifest version that is requested?
Is it that the server _should_ give back a manifest serialized with this version, or that it _can_ give back a manifest up to the specified version?
In case of the former an error should be returned when requesting a newer version than known by the server, in case of the latter the code here should read `(min 2 new_version)`.
As it is right now a newer (as of yet unexisting client) will have problems when explaining to an old server that it wants manifests in version format 3.

(My preference goes towards the "it can give back a manifest up to the specified version" semantics.)

cpp/src/lib/manifest.cc
```diff
@@ -205,22 +202,127 @@ template <> void from2(message &m, Manifest &mf, bool &ok_to_continue) {
...
+  for (int32_t c = n_chunks - 1; c >= 0; --c) {
+    uint32_t n_fragments;
+    from(m2, n_fragments);
+    std::vector<std::shared_ptr<Fragment>> chunk(n_fragments);
+    for (int32_t f = n_fragments - 1; f >= 0; --f) {
+      std::shared_ptr<Fragment> fragment_ptr(new Fragment());
+      from(m2, *fragment_ptr);
+      chunk[f] = std::move(fragment_ptr);
+    };
+    mf.fragments[c] = std::move(chunk);
```
toolslive: yes: we need a test.

ocaml/src/proxy_server.ml
```diff
@@ -596,6 +607,27 @@ let proxy_protocol (alba_client : Alba_client.alba_client)
...
+               begin
+                 match vo with
+                 | None -> ()
+                 | Some v ->
+                    let new_version = deserialize Llio.int8_from v in
+                    ProxySession.set_manifest_ser session new_version
+               end
+            | key,_ ->
+               let open Nsm_model in
+               Err.failwith Err.Session_update_error ~payload:key
```
domsj: So for unknown keys this throws an exception...
That makes it harder for future clients that want to specify multiple keys (in 1 call), as they might talk to an old server which only supports a subset of the keys. So they would have to fall back to multiple UpdateSession calls anyway (then why take a list in the first place).
Maybe it makes more sense here to return the keys that the server was able to interpret?

(Similar remark for nsm host session keys)

cpp/src/lib/rora_proxy_client.h
```diff
@@ -117,6 +117,12 @@ class RoraProxy_client : public Proxy_client {
   string _fragment_key(const namespace_t namespace_id, const string &object_id,
                        uint32_t version_id, uint32_t chunk_id,
                        uint32_t fragment_id);
+  boost::optional<int> _ser_version;
```
domsj: this is no longer used and can thus be removed?

cpp/src/lib/manifest.cc
toolslive: I don't think (obviously backed by the test) it's the wrong order.
https://github.com/openvstorage/alba/blob/master/cpp/include/llio.h#L262

cpp/src/lib/rora_proxy_client.h
toolslive: we might want to know... (for example in a test)

cpp/include/generic_proxy_client.h
```diff
@@ -69,6 +69,12 @@ class GenericProxy_client : public Proxy_client {
                        const consistent_read,
                        std::vector<proxy_protocol::object_info> &,
                        alba::statistics::RoraCounter &);
+  virtual void
+  read_objects_slices3(const std::string &namespace_,
```
toolslive: yes.

cpp/src/lib/proxy_protocol.cc
```diff
@@ -334,13 +353,20 @@ void read_read_objects_slices2_response(
...
+}
+
 void write_apply_sequence_request(
     message_builder &mb, const string &namespace_, const bool write_barrier,
     const std::vector<std::shared_ptr<alba::proxy_client::sequences::Assert>>
         &asserts,
     const std::vector<std::shared_ptr<alba::proxy_client::sequences::Update>>
         &updates) {
-  write_tag(mb, _APPLY_SEQUENCE);
+  write_tag(mb, _APPLY_SEQUENCE2);
```
toolslive: indeed. suppressed in favour of session based concept.

cpp/src/lib/manifest.cc
```diff
@@ -205,22 +202,127 @@ template <> void from2(message &m, Manifest &mf, bool &ok_to_continue) {
...
+  from(m2, mf.compression);
+  from(m2, mf.encrypt_info);
+  from(m2, mf.checksum);
+  from(m2, mf.size);
+  uint8_t layout_tag;
+  from(m2, layout_tag);
+  if (layout_tag != 1) {
+    throw deserialisation_exception("unexpected layout tag");
+  }
+  // from(m2, mf.fragments);
```
toolslive: I'll phrase it as a todo. I actually want the templating system to do this automagically for me.

ocaml/src/alba.ml
```diff
@@ -435,23 +435,31 @@ let alba_show_object
        >>= fun (hm,r) ->
        match r with
          | None ->
-            Lwt_io.eprintlf "not found"
+            Lwt.fail_with "not found"
```
domsj: this path doesn't take the to-json flag into account...
Otoh if we add a note about the to-json flag for this command being for dev/test purposes only then it doesn't have to

ocaml/src/nsm_model.ml
```diff
@@ -1132,6 +1112,7 @@ module Err = struct
     | Assert_failed           [@value 20]
     | Invalid_bucket          [@value 21]
     | Preset_violated         [@value 22]
+    | Session_update_error    [@value 23]
```
domsj: this isn't used anywhere (anymore), so could be removed

ocaml/src/nsm_host_plugin.ml
```diff
@@ -375,6 +378,30 @@ let handle_query : type i o. read_user_db -> (i, o) Nsm_host_protocol.Protocol.q
...
+             | "manifest_ser" ->
+                begin
+                  match vo with
+                  | None -> acc
+                  | Some v ->
+                     let wanted_version = deserialize Llio.int8_from v in
+                     let new_version = min 2 wanted_version in
+                     let () =
+                       Nsm_protocol.Session.set_manifest_ser session new_version in
+                     (k, v) :: acc
```
domsj: server side should return the actual new_version that it will use to serialize manifests? (and that can be used by the client)

ocaml/src/nsm_model.ml
```diff
@@ -1598,7 +1579,9 @@ module NamespaceManager(C : Constants)(KV : Read_key_value_store) = struct
       Update'.compare_and_swap
         (Keys.objects ~object_id)
         None
-        (Some (serialize Manifest.output manifest)) in
+        (Some (serialize (Manifest.to_buffer
+                            ~version:1 (* TODO: should become 2 *)) manifest))
```
domsj: "TODO: should become 2" ... once all nsm nodes have upgraded that is? I guess the TODO can stay then for now.

ocaml/src/proxy_client.ml
```diff
@@ -75,16 +75,19 @@ class proxy_client fd =
 
     with_response tag_name response_deserializer f
   in
+  let session = ProxySession.make () in
```
domsj: discover available manifest version for this proxy client too?

ocaml/src/proxy_server.ml
```diff
@@ -596,6 +607,29 @@ let proxy_protocol (alba_client : Alba_client.alba_client)
...
+               match k with
+               | "manifest_ser" ->
+                  begin
+                    match vo with
+                    | None -> acc
+                    | Some v ->
+                       let wanted_version = deserialize Llio.int8_from v in
+                       let new_version = min 2 wanted_version in
+                       let () = ProxySession.set_manifest_ser session new_version in
+                       (k, v) :: acc
```
domsj: server side should return the actual new_version that it will use to serialize manifests? (and that can be used by the client)

ocaml/src/alba.ml
toolslive: right. we should reserve the right to do whatever we want to do with the json format

toolslive: but, it does take the --to-json into account, throwing results in something like
```
{"success":false,
  "error":{ "message":"(Failure \"not found\")", 
            "exception_type":"unknown",
            "exception_code":0
          }
}
```

ocaml/src/nsm_model.ml
toolslive: ack.

ocaml/src/nsm_model.ml
toolslive: We also might want to bulk change stored manifests as well at some point. I'll leave the TODO.

ocaml/src/nsm_host_plugin.ml
toolslive: ack.

ocaml/src/proxy_server.ml
toolslive: ack.

ocaml/src/proxy_client.ml
toolslive: ack.

ocaml/src/proxy_client.ml
```diff
@@ -226,8 +230,26 @@ let make_client ip port transport =
...
-    (fun () -> _prologue nfd Protocol.magic Protocol.version >>= fun () ->
-               Lwt.return (new proxy_client nfd, closer))
+    (fun () ->
+      _prologue nfd Protocol.magic Protocol.version >>= fun () ->
+      let session = ProxySession.make () in
+      let client = new proxy_client nfd session in
+      let manifest_ser = 2 in
+      let kvs = [("manifest_ser", Some (serialize Llio.int8_to manifest_ser))]
+      in
+      client # update_session kvs >>= fun processed ->
```
domsj: This might fail with UnknownOperation on older proxies, so for backwards compatibility that error should be caught and ignored

toolslive: Jenkins beat you to the punch ;)

PR #628, opened 2017-03-01
Fnr

toolslive:
Note: 
- this will do absolutely nothing until the manifest serialization version is 2 in the NSM.
- enabling this (along with infrastructure to trigger it at the right moment) will be done in a different pr.

### DIFF
--- a/ocaml/src/alba_client_upload.ml
+++ b/ocaml/src/alba_client_upload.ml
@@ -218,8 +218,10 @@ let upload_chunk
            in
 
            let res =
-             let loc = osd_id_o, version_id in
-             Fragment.make loc checksum packed_len fragment_ctr fnro
+             Fragment.make
+               osd_id_o version_id
+               checksum packed_len fragment_ctr
+               fnro
            in
            Lwt_log.debug_f "fragment_uploaded (%i,%i) to %s"
                            chunk_id fragment_id ([%show : int64 option] osd_id_o)
@@ -238,7 +240,7 @@ let upload_chunk
            Lwt.return_unit
          )
      in
-     let test = fun (_, f) -> Fragment.has_loc f in
+     let test = fun (_, f) -> Fragment.has_osd f in
      Lwt_extra2.first_n
        ~count:min_fragment_count
        ~slack:upload_slack
@@ -504,8 +506,7 @@ let upload_object''
                  })
                in
                let fragment =
-                 let loc = None, version_id in
-                 Fragment.make loc
+                 Fragment.make None version_id
                           fragment_checksum
                           packed_fragment_size
                           fragment_ctr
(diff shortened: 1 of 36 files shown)


ocaml/src/osd.ml
```diff
@@ -45,9 +45,16 @@ module Osd_namespace_state = struct
 end
 
 module Error = Asd_protocol.Protocol.Error
+
+type apply_result' = ((key * string option) list) [@@deriving show]
+
 type apply_result =
```
domsj: hmm, looking at this type we may want use Prelude.Result instead?

ocaml/src/tools/prelude.ml
```diff
@@ -191,6 +191,16 @@ module List = struct
 
   let find_index f l = find_index' f 0 l
 
+  let assoc' eq k kvs =
+    let rec inner = function
+      | [] -> raise Not_found
```
toolslive: ok.

ocaml/src/alba_client_upload.ml
```diff
@@ -107,7 +113,19 @@ let upload_chunk
       ~osds
       ~(fragment_cache : Fragment_cache.cache)
       ~cache_on_write
-      ~upload_slack
+      ~upload_slack :
+      ((Alba_statistics.Statistics.fragment_upload
+        * (Albamgr_protocol.Protocol.Osd.id option
+           * Asd_protocol.checksum
+           * bytes option
+           * string option
```
toolslive: ok.

ocaml/src/asd_server.ml
```diff
@@ -1222,12 +1222,32 @@ let execute_update : type req res.
...
+                       (fun acc ku ->
+                         match ku with
+                         | (key, `Set (value, _)) ->
+                            begin
+                              let open Value in
+                              let e =
+                                match value with
+                                | Direct v -> key, None
+                                | OnFs (fnr, _) ->
+                                   let fnr_s = serialize Llio.int64_to fnr in
```
toolslive: ok. It could also be a varint in rocksdb. (it's viral)

ocaml/src/osd.ml
toolslive: what about Pervasives.result ?

domsj: That's actually the same type if I'm not mistaken?

ocaml/src/nsm_client.ml
```diff
@@ -169,32 +171,55 @@ class client (nsm_host_client : Nsm_host_client.basic_client) namespace_id =
...
-            | Nsm_exn (Unknown_operation,_) ->
-               supports_update_manifest2 := Some false;
-               old_call ()
-            | exn -> Lwt.fail exn
-           )
-
+      | Some 1 -> old_call ()
+      | None ->
+         begin
+           let call = function
```
domsj: some repetition here matching on 3, 2, 1 that could be avoided

PR #652, opened 2017-03-15
upload from buffer in sequence

toolslive:


domsj:
looks fine to me, maybe @redlicha would have some comments?

toolslive:
`const std::shared_ptr<std::vector<char>>` for the blob might be overkill compared to `const std::vector<char>&`

cpp/include/proxy_sequences.h
```diff
@@ -118,7 +118,30 @@ class UpdateUploadObjectFromFile final : public Update {
   const alba::Checksum *_cs_o;
 };
 
-/* TODO upload object from in memory blob */
+class UpdateUploadObject final : public Update {
+public:
+  UpdateUploadObject(const std::string &name,
+                     const std::shared_ptr<std::vector<char>> data,
```
redlicha: The `shared_ptr` could go away as `mb.add_raw` below has a comment that it copies. Which leaves the vector - callers might not have the data in a vector, so a plain `const uint8_t` together with a `size_t` would be more generic.

Excursion: the CPP core guidelines / its support lib come with a `span` template:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#bounds1-dont-use-pointer-arithmetic-use-span-instead
https://github.com/Microsoft/GSL

Using the GSL for voldrv is eventually is on my wishlist, but that needs a more recent compiler than available to us at the moment (on Ubuntu 14.04).

cpp/include/proxy_sequences.h
```diff
@@ -174,6 +197,13 @@ class Sequence {
     return *this;
   }
 
+  Sequence &add_upload(const std::string &name,
+                       const std::shared_ptr<std::vector<char>> blob,
+                       const alba::Checksum *cs_o) {
+    _updates.push_back(
+        std::shared_ptr<Update>(new UpdateUploadObject(name, blob, cs_o)));
```
redlicha: using `std::make_shared<UpdateUploadObject>(name, blob, cs_o)` is preferrable (also applies to the existing calls):
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r22-use-make_shared-to-make-shared_ptrs

cpp/include/proxy_sequences.h
```diff
@@ -118,7 +118,30 @@ class UpdateUploadObjectFromFile final : public Update {
...
+      : _name(name), _data(data), _cs_o(cs_o){};
+
+  void to(llio::message_builder &mb) const override {
+    mb.add_type(2);
+    llio::to(mb, _name);
+    uint32_t len = _data->size();
+    llio::to(mb, len);
+    mb.add_raw(_data->data(), len); // copies
+    if (_cs_o == nullptr) {
+      llio::to<boost::optional<const Checksum *>>(mb, boost::none);
```
redlicha: Why is the ptr wrapped into an optional - is `boost::none` handled differently from a `boost::optional<const CheckSum*>(nullptr)`?

cpp/include/proxy_sequences.h
toolslive: yes, a pointer to the blob + size is most generic, but the copy might happen way later, so you need to keep the buffer alive until it happens.

cpp/include/proxy_sequences.h
toolslive: the problem is that `boost::none == boost::optional<X>(nullptr)` yields `false` so the serializer will try to deref the optional and SEGV in
```
template <typename X>
void to(message_builder &mb, const boost::optional<X> &xo) noexcept {
  if (boost::none == xo) {
    to(mb, false);
  } else {
    to(mb, true);
    to(mb, *xo);
  }
}
```

PR #662, opened 2017-03-21
catch another exception for race between namespace deletion & get-disk-safety

### LINKED ISSUE #657: get-disk-safety failed with 'Namespace manager exception: Nsm_model.Err.Namespace_id_not_found'.
### Problem description
Monitoring revealed the following sub awesome behavior with the get disk safety command:
```
CRIT - EXCEPTION HC000 - Could not fetch alba information for backend nvmebackend Message: Command 'get-disk-safety' failed with 'Namespace manager exception: Nsm_model.Err.Namespace_id_not_found'.
```

What could have happened:
- Another healtcheck was busy with a test that involves creating and removing namespaces
- At the time the other namespace was getting deleted, get-disk-safety was called

### Proposed solution
The whole command should not fail when one namespace cannot be fetched. Perhaps return the current output you have collected and add an exception section or something?

domsj:
fixes #657

### DIFF
--- a/ocaml/src/disk_safety.ml
+++ b/ocaml/src/disk_safety.ml
@@ -109,7 +109,11 @@ let get_disk_safety alba_client namespaces dead_osds =
             (get_dead_namespace_osds ~namespace_id) >>= fun r ->
           Lwt.return (Some (namespace, r)))
         (function
-         | Error.Albamgr_exn (Error.Namespace_does_not_exist, _) -> Lwt.return_none
+         | Nsm_model.Err.Nsm_exn (Nsm_model.Err.Namespace_id_not_found, _)
+         | Error.Albamgr_exn (Error.Namespace_does_not_exist, _) ->
+            (* getting disk-safety might race with namespace deletion.
+             * we want to ignore errors caused by this race. *)
+            Lwt.return_none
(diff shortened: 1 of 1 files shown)


ocaml/src/disk_safety.ml
```diff
@@ -109,7 +109,11 @@ let get_disk_safety alba_client namespaces dead_osds =
             (get_dead_namespace_osds ~namespace_id) >>= fun r ->
           Lwt.return (Some (namespace, r)))
         (function
-         | Error.Albamgr_exn (Error.Namespace_does_not_exist, _) -> Lwt.return_none
+         | Nsm_model.Err.Nsm_exn (Nsm_model.Err.Namespace_id_not_found, _)
+         | Error.Albamgr_exn (Error.Namespace_does_not_exist, _) ->
+            (* getting disk-safety might race with namespace deletion.
```
JeffreyDevloo: Could you provide a log here so we can catch it somewhere and log it in our albacli wrapper.
Throwing away the info in this thread would seem a waste.
(for instance what namespace was not available)

domsj: Ok, added some logging

Issue #660, opened 2017-03-20
partial reads: sub awesome behaviour of concurrent partial reads for same fragment


toolslive:
These result in the same fragment being read and added to the fragment cache multiple times.

domsj:
(should be) fixed by #665 & #666

PR #665, opened 2017-03-21
deduplicate fragment fetch

domsj:
part of changes required for #660


domsj:
you're right, added a test for this behaviour

PR #670, opened 2017-03-23
log on info level when making a new connection

domsj:
There shouldn't be too many of these. And if somehow there are, we want it to be visible

### DIFF
--- a/ocaml/src/albamgr_client.ml
+++ b/ocaml/src/albamgr_client.ml
@@ -745,7 +745,7 @@ let retrieve_cfg_from_any_node ~tls_config current_config =
 let make_client tls_config buffer_pool (ccfg:Arakoon_client_config.t) =
   let open Client_helper in
   let tls = Client_helper.get_tls_from_ssl_cfg (Option.map Tls.to_ssl_cfg tls_config) in
-  Lwt_log.debug_f "Albamgr_client.make_client :%s " (Arakoon_client_config.show ccfg)
+  Lwt_log.info_f "Albamgr_client.make_client :%s " (Arakoon_client_config.show ccfg)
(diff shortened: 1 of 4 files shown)


ocaml/src/tools/net_fd.ml
```diff
@@ -130,7 +130,15 @@ let apply_keepalive tcp_keepalive = function
      Tcp_keepalive.apply fd tcp_keepalive
   | Rsocket fd -> ()
 
-let connect fd address = match fd with
+let show_sockaddr =
```
toolslive: there's also Network.a2s

ocaml/src/albamgr_client.ml
```diff
@@ -745,7 +745,7 @@ let retrieve_cfg_from_any_node ~tls_config current_config =
 let make_client tls_config buffer_pool (ccfg:Arakoon_client_config.t) =
   let open Client_helper in
   let tls = Client_helper.get_tls_from_ssl_cfg (Option.map Tls.to_ssl_cfg tls_config) in
-  Lwt_log.debug_f "Albamgr_client.make_client :%s " (Arakoon_client_config.show ccfg)
+  Lwt_log.info_f "Albamgr_client.make_client :%s " (Arakoon_client_config.show ccfg)
```
toolslive: do we want to show the whole config or merely the cluster id ?

ocaml/src/tools/net_fd.ml
domsj: ack

ocaml/src/albamgr_client.ml
domsj: good point, the entire thing is a bit too verbose

PR #676, opened 2017-03-24
quick fix for master

toolslive:


ocaml/src/remotes.ml
```diff
@@ -99,8 +115,21 @@ module Pool = struct
               t.pool_size
               ~check:(fun _ ->
                 function
-                | Nsm_model.Err.Nsm_exn _ ->
-                   true
+                | Nsm_model.Err.Nsm_exn (t,p) ->
+                   begin
+                     let r =
+                       match t with
+                       | Nsm_model.Err.Inconsistent_read -> false
```
domsj: add `Not_master` and `Unknown` here too?

toolslive: yes.

PR #683, opened 2017-03-27
sometimes you want to drop the entire pool

toolslive:


ocaml/src/tools/lwt_pool2.ml
```diff
@@ -81,18 +86,35 @@ let use t f =
...
+                end
+              | DropPool ->
+                 begin
+                   t.count <- 0;
+                   t.cleanup a >>= fun () ->
+                   let rec loop () =
+                     if Queue.is_empty t.available_items
+                     then Lwt.return_unit
+                     else
+                       let a = Queue.pop t.available_items in
```
domsj: Preferably pop all items from queue without lwt binds in between?
Currently it's hard to reason about what will happen when another thread wants to use the pool after count has been reset but before all connections have been popped from the queue...

ocaml/src/albamgr_client.ml
```diff
@@ -614,49 +614,71 @@ object(self)
...
+  do_request
+    connection
+    (command_to_tag (Wrap_q command))
+    (fun buf ->
+      Consistency.to_buffer buf consistency;
+      write_query_i command buf req)
+    (read_query_o command)
+
+let _get_version connection =
+  _query connection ~consistency:Consistency.Consistent GetVersion ()
```
domsj: I think this read should be done with consistency `No_guarantees`

ocaml/src/nsm_host_client.ml
```diff
@@ -30,62 +30,85 @@ class type basic_client = object
...
+        connection
         session
         (command_to_tag (Wrap_q command))
         (fun buf ->
          Consistency.to_buffer buf consistency;
          write_query_i command buf req)
         (read_query_o command)
 
+let _get_version connection session =
+  _query connection session ~consistency:Consistency.Consistent GetVersion ()
```
domsj: Same remark as in albamgr_client: I think this can (and should) be an inconsistent read

domsj:
We need to make sure to run all tests again once 1.3.11 has been tagged and before pushing the binary towards the package repos, as putting the tag will change the behavior that is triggered...

PR #697, opened 2017-04-05
support for partial reads from c++ rora client when using (ctr mode) encryption

### LINKED ISSUE #447: Rora support for CTR mode encryption
https://github.com/openvstorage/alba/pull/444 introduces CTR mode for encryption, which is compatible with partial reads. The rora path could/should have support for this too (currently it will/should do reads through the proxy as a fallback).

domsj:
Fixes #447.
This PR also fixes a bug in the varint serialisation.


cpp/src/lib/rora_proxy_client.cc
```diff
@@ -451,6 +459,42 @@ void RoraProxy_client::read_objects_slices(
...
+              ALBA_LOG(
+                  INFO,
+                  "Could not partially decrypt data, which is unexpected!");
+              throw 0;
+            }
+            break;
+          }
+        }
+      } catch (...) {
+        result_front = -1;
```
toolslive: I think you want to log (at least debug level), what got thrown

cpp/include/generic_proxy_client.h
```diff
@@ -112,6 +112,10 @@ class GenericProxy_client : public Proxy_client {
 
   virtual bool has_local_fragment_cache();
 
+  virtual boost::optional<string>
+  get_fragment_encryption_key(const string &alba_id,
```
toolslive: why the `_fragment_` in the name ?

domsj: I named it like this because the property in the preset is named "fragment_encryption" ... but I'm not attached to the name

cpp/src/lib/rora_proxy_client.cc
domsj: ok

PR #702, opened 2017-04-24
Grammatical Issue in bash

rainmanh:
Bash Issue in the jenkins bash script.


domsj:
Thanks for the contribution @rainmanh

PR #707, opened 2017-04-27
Log enhancements

dejonghb:
As @domsj predicted in  #705....

### DIFF
--- a/ocaml/src/alba_client_download.ml
+++ b/ocaml/src/alba_client_download.ml
@@ -119,9 +119,10 @@ let download_packed_fragment
   | Some data ->
      osd_access # get_osd_info ~osd_id >>= fun (_, state,_) ->
      Osd_state.add_read state;
-     Lwt_log.ign_debug_f
+     Lwt_log.debug_f
        ">>> download_packed_fragment: object (%S, %S) chunk %i, fragment %i from osd_id:%Li -- ENDS %i bytes @ %nX"
-       object_name object_id chunk_id fragment_id osd_id (Lwt_bytes.length data) (Lwt_bytes.raw_address data);
+       object_name object_id chunk_id fragment_id osd_id (Lwt_bytes.length data) (Lwt_bytes.raw_address data)
+     >>= fun () ->
      E.return (osd_id, data)
 
 (* consumers of this method are responsible for freeing
--- a/ocaml/src/alba_client_download.ml
+++ b/ocaml/src/alba_client_download.ml
@@ -77,8 +77,8 @@ let download_packed_fragment
   let osd_id, version_id = location in
 
   Lwt_log.debug_f
-    "download_packed_fragment: object (%S, %S) chunk %i, fragment %i from osd_id:%Li"
-    object_id object_name
+    "download_packed_fragment: object (%S, %S) chunk %i, fragment %i from osd_id:%Li STARTS"
+    object_name object_id
     chunk_id fragment_id
     osd_id
   >>= fun () ->
@@ -112,10 +112,16 @@ let download_packed_fragment
     )
   >>== function
   | None ->
+     Lwt_log.ign_debug_f
+       ">>> download_packed_fragment: object (%S, %S) chunk %i, fragment %i from osd_id:%Li -- ENDS returning FragmentMissing <<<"
+       object_name object_id chunk_id fragment_id osd_id;
      E.fail `FragmentMissing
   | Some data ->
      osd_access # get_osd_info ~osd_id >>= fun (_, state,_) ->
      Osd_state.add_read state;
+     Lwt_log.ign_debug_f
+       ">>> download_packed_fragment: object (%S, %S) chunk %i, fragment %i from osd_id:%Li -- ENDS %i bytes @ %nX"
+       object_name object_id chunk_id fragment_id osd_id (Lwt_bytes.length data) (Lwt_bytes.raw_address data);
      E.return (osd_id, data)
 
 (* consumers of this method are responsible for freeing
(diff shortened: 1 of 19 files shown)


ocaml/src/alba_client_download.ml
```diff
@@ -112,10 +112,16 @@ let download_packed_fragment
...
   >>== function
   | None ->
+     Lwt_log.ign_debug_f
+       ">>> download_packed_fragment: object (%S, %S) chunk %i, fragment %i from osd_id:%Li -- ENDS returning FragmentMissing <<<"
+       object_name object_id chunk_id fragment_id osd_id;
      E.fail `FragmentMissing
   | Some data ->
      osd_access # get_osd_info ~osd_id >>= fun (_, state,_) ->
      Osd_state.add_read state;
+     Lwt_log.ign_debug_f
```
toolslive: why `ign_` ? This is already in between binds.

ocaml/src/asd_server.ml
```diff
@@ -394,7 +394,7 @@ module Net_fd = struct
        in
        let writer buffer offset length =
          let write_from_source = Lwt_ssl.write_bytes socket buffer in
-         Lwt_extra2._write_all write_from_source offset length
+         Lwt_extra2._write_all write_from_source (identifier fd_out) offset length
```
toolslive: This converts the fd to an int (and on the callee side back into an fd). 
It's not necessary and drops any type safety we have.

ocaml/src/osd_access.ml
```diff
@@ -158,8 +158,18 @@ module Osd_pool = struct
               let open Lwt_pool2 in
               match exn with
               | Asd_protocol.Protocol.Error.Exn _ -> Keep
+              | Unix.Unix_error(Unix.EFAULT, _, _) ->
+                 Lwt_log.ign_info_f
+                   ~exn
```
toolslive: drop `~exn` (it doesn't really work and you put the callstack in the message anyway)

ocaml/src/other/posix_stubs.c
```diff
@@ -62,9 +57,7 @@ void _bs_posix_fadvise(value fd, value offset, value len, value advice) {
                         break;
         }
 
-        enter_blocking_section();
```
toolslive: why was the sandwich removed?

ocaml/src/tools/hashes.ml
```diff
@@ -86,7 +86,8 @@ class crc32c_hasher =
         self # update_substring s 0 (String.length s)
 
       method update_lwt_bytes lwt_bytes pos len =
-        cur <- Alba_crc32c.Crc32c.big_array ~crc:cur lwt_bytes pos len false
+        if len > 0
```
toolslive: The test was also added on the inside of the crc32c function ?

ocaml/src/tools/lwt_bytes2.ml.cppo
```diff
@@ -29,14 +29,48 @@ module Lwt_bytes = struct
   let pp formatter t =
     Format.pp_print_string formatter (show t)
 
-  let unsafe_destroy (t : t) = Core_kernel.Bigstring.unsafe_destroy t
+  let raw_address (t : t) =
+    let open Ctypes in
+    let x = bigarray_start array1 t in
+    raw_address_of_ptr (to_voidp x)
+
+  let unsafe_destroy ?(msg=(Printexc.get_callstack 5 |> Printexc.raw_backtrace_to_string)) (t : t) =
```
toolslive: performance considerations for the evaluation of msg?

ocaml/src/asd_server.ml
domsj: the callee side does not turn it back into an fd

ocaml/src/tools/lwt_bytes2.ml.cppo
domsj: that's an interesting point.
@dejonghb it probably should be written as
```
let unsafe_destroy ?msg (t : t) =
  ...
  Lwt_log.ign_debug_f "..." l address
                         (match msg with
                           | Some msg -> msg
                           | None -> Printexc.get_callstack 5 |> Printexc.raw_backtrace_to_string);
  ...
```

ocaml/src/other/posix_stubs.c
dejonghb: I don't expect the posix_fadvise to take long enough to have any benifit of releasing/reacquiring the runtime_system, but I could be wrong here...

ocaml/src/tools/hashes.ml
dejonghb: Having the test here does not lead into the stubs part... (while running maintenance tests zero length arrays were passed in quite often)

ocaml/src/tools/alba_wrappers_stubs.c
```diff
@@ -36,9 +36,10 @@ but WITHOUT ANY WARRANTY of any kind.
    would make sense.
 */
 CAMLprim value alba_get_maxrss(value unit) {
-    CAMLparam1 (unit);
+    CAMLparam1(unit);
+    CAMLlocal1(vres);
 
-    caml_release_runtime_system();
+    //caml_release_runtime_system();
```
domsj: if getrusage is considered to be a cheap call we should probably also remove Alba_wrappers.Sys2.lwt_get_maxrss?

ocaml/src/asd_server.ml
toolslive: sorry, was too eager there.

ocaml/src/tools/alba_wrappers_stubs.c
toolslive: I don't think getrusage is cheap. I might be wrong too ;)

PR #709, opened 2017-05-04
bash completion (first stab)

toolslive:
[asciicast](https://asciinema.org/a/0frfftd084lzq5ilo1oqd393s)

### DIFF
--- a/ocaml/src/alba.ml
+++ b/ocaml/src/alba.ml
@@ -952,6 +952,18 @@ let () =
         Asd_bench.cmds;
         cmds1; ]
   in
-  match Term.eval_choice default_cmd cmds with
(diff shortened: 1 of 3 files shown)


ocaml/src/alba.ml
```diff
@@ -952,6 +952,18 @@ let () =
...
         cmds1; ]
   in
-  match Term.eval_choice default_cmd cmds with
+  let cmd_names =
+    "commands"
+    :: List.map
+         (fun (_,info) -> Term.name info )
+         cmds
+  in
+  let names verbose = Printf.printf "%s\n" (String.concat " " cmd_names) in
```
domsj: remove unused `verbose` argument?

PR #715, opened 2017-05-10
also remove ppx binary on clean

toolslive:


Makefile
```diff
@@ -10,7 +10,8 @@ clean:
 	cd setup && ocamlbuild -clean
```
domsj: also add a `cd ppx && ocamlbuild -clean` here?

toolslive: ok.

PR #717, opened 2017-05-11
Robustify stress2 test + use tail recursive map everywhere

domsj:
(all .ml files are as of now required to `open! Prelude`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,11 @@ build-cmxs: build-alba
 
 build-nsm-plugin: build-cmxs
 	cd ocaml && ocamlfind ocamlopt \
+	_build/src/tools/prelude.cmx \
 	_build/alba_version.cmx \
 	_build/src/range_query_args.cmx \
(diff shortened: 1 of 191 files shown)


aantron:
Let me know if you do decide that tail-recursive `Lwt_list` functions (https://github.com/ocsigen/lwt/pull/347) are necessary. We can quickly add them. You should be able to pin to that Lwt commit, and it will also get released in the next release. We'll probably eventually switch to tail-recursive functions no matter what, once it's time to deal with "everything" in `Lwt_list`, which shouldn't be too far off.

domsj:
We have included our own version of the Lwt_list functions ... see https://github.com/openvstorage/alba/blob/1.3.15/ocaml/src/tools/prelude.ml#L856, so no rush for a new release from our end.
As an fyi: the stress2 suite still periodically segfaults on my machine. But as it's only on my machine I'm starting to think it's a hardware problem.

PR #726, opened 2017-05-29
log when negative disk usage detected. #209

toolslive:


ocaml/src/asd_server.ml
```diff
@@ -1249,7 +1251,13 @@ let execute_update : type req res.
                   (function
                     | Lwt.Canceled -> Lwt.fail Lwt.Canceled
                     | ConcurrentModification -> Lwt.return `Retry
-                    | exn -> Lwt.fail exn) >>= function
+                    | NegDiskUsage ->
+                       Lwt_log.fatal "negative disk usage detected." >>= fun () ->
+                       Lwt_extra2.flush_logging () >>= fun () ->
+                       let () = exit 666 in
+                       Lwt.return `Retry
```
domsj: `exit 666` can just be the last statement .. doesn't the compiler warn this is dead code?

toolslive: unification goes outside in, so the `Lwt.return ...` (how to escape that backtick in inline markdown?) is needed. `exit` is a regular function with an irregular result. In short: no warning.

domsj: Fyi: No, regardless of how unification works, the `Lwt.return ...` is not needed. The warning isn't there because you wrapped `exit 666` in `let () = ... in`

PR #730, opened 2017-05-31
consider removal of broken objects

toolslive:


ocaml/src/maintenance_helper.ml
```diff
@@ -285,3 +285,126 @@ let upload_missing_fragments
...
+  | Rebalance
+  | Regenerate
+  | Rewrite
+[@@deriving show]
+
+let compare_buckets (bucket1, j1) (bucket2,j2) =
+  let compare_bucket_safety
+        (k1, _, fragment_count1, _)
+        (k2, _, fragment_count2, _) =
+    (fragment_count2 - k2) - (fragment_count1 - k1)
```
domsj: you flipped this around which is a bit counter intuitive?

ocaml/src/maintenance_test.ml
```diff
@@ -1016,7 +1016,61 @@ let test_repair_evolved_compressor () =
...
+      best_actual_max_disks_per_node
+      policies
+      is_cache_namespace
+      bucket_count
+  in
+
+  let expected =[
+      ( 1, 2,  3, 1), Rewrite;     (* safety: 2 *)
+      (16, 8, 20, 3), Regenerate;  (* safety: 4 *)
+      (16, 8, 24, 6), Rebalance;
```
domsj: can you add 1 extra Rewrite & Regenerate case so that part of the sorting is fully tested too?

ocaml/src/maintenance.ml
```diff
@@ -1167,88 +1167,43 @@ class client ?(retry_timeout = 60.)
...
+          (fun prefix preset acc ->
+            acc ||
+              (let open Albamgr_protocol.Protocol in
+                preset = namespace.Namespace.preset_name
+               &&
+                 let len = String.length namespace_name in
+                 let prefix_len = String.length prefix in
+                 len >= prefix_len
+                 &&
+                   String.sub namespace_name 0 prefix_len = prefix
```
toolslive: I had that feeling too, looked for it, couldn't find it. Will add to prelude.

ocaml/src/maintenance_helper.ml
```diff
@@ -285,3 +285,126 @@ let upload_missing_fragments
...
+       end
+    | Rewrite ->
+       begin
+         match j2 with
+         | ConsiderRemoval | Rebalance -> 1
+         | Regenerate | Rewrite -> compare_bucket_safety bucket1 bucket2
+       end
+  end
+
+let neg_compare_buckets b1 b2 =
```
toolslive: ok.

ocaml/src/maintenance_test.ml
toolslive: ok.

ocaml/src/maintenance.ml
```diff
@@ -1167,88 +1167,43 @@ class client ?(retry_timeout = 60.)
...
-      let (_, bucket_count) = stats.Nsm_model.NamespaceStats.bucket_count in
+      let open Nsm_model in
+      let (_, bucket_count) = stats.NamespaceStats.bucket_count in
 
       Lwt_log.debug_f
         "Found buckets %s for namespace_id:%Li"
         ([%show : (Policy.policy * int64) list] bucket_count)
         namespace_id >>= fun () ->
 
+      mgr_access # get_namespace_by_id ~namespace_id
```
toolslive: that's silly and easy to avoid as we have the name a level above. will do.

ocaml/src/tools/prelude.ml
```diff
@@ -101,6 +101,16 @@ end
 module String = struct
   include String
 
+  let starts_with s prefix =
```
domsj: This implementation is inefficient due to bounds checks, and it can throw `Invalid_argument 'index out of bounds'` when the prefix is longer than the string s.
Probably the easiest solution is to use `Core.String.is_prefix ~prefix s`

PR #757, opened 2017-07-17
list-work in multiple steps iso one

toolslive:


ocaml/src/cli_mgr.ml
```diff
@@ -533,34 +533,74 @@ let alba_list_participants_cmd =
...
+        let max = if max = -1 then max_int else max in
+        let rec loop acc first finc more cnt =
+          match (more, cnt) with
+          | (false, _ ) -> Lwt.return ((cnt, acc), false)
+          | (true, cnt) ->
+             if cnt = max then Lwt.return ((cnt, acc), true)
+             else
+               begin
+                 let todo = max - cnt in
+                 let batch =
```
domsj: `let batch = min todo batch_size in`  ?

toolslive: ack

PR #762, opened 2017-07-19
More efficient way to fetch the work item count

domsj:


ocaml/src/arith64.ml
```diff
@@ -14,14 +14,15 @@ let operation_of = function
...
-let make_update key operation x=
+let make_update key operation x ~clear_if_zero =
   match operation with
   | PLUS_EQ ->
      begin
      let buf = Buffer.create 20 in
      Llio.int8_to buf 1;
      Llio.string_to buf key;
      Llio.int64_to buf x;
+     Llio.bool_to buf clear_if_zero;
```
toolslive: why this ?

domsj: this is to ensure the workitemcount key remains available on count 0 such that ensure_work_item_count doesn't have to recount the (zero) work items

domsj:
I forgot to add a code for the newly added operation. Saw this during manual verification, and also noticed that the albamgr doesn't handle unknown operations correctly (despite our attempt to test this!).
So expect another commit soon.

PR #788, opened 2017-08-16
there are no parity replicas

toolslive:


ocaml/src/alba_client_download.ml
```diff
@@ -132,6 +132,7 @@ let download_fragment
 
   let t0_fragment = Unix.gettimeofday () in
 
+  let fragment_id = if k = 1 then 0 else fragment_id in
```
domsj: This also changes the value passed on to `download_packed_fragment`, which in effect could result in ignoring the configured read preference?
I think you only want to ignore the fragment_id for the cache_key and dedup_key bindings in this function?

toolslive: indeed.

Issue #777, opened 2017-08-08
proxy should also refresh & update abm_cfg of fragment_cache backend


domsj:
similar to how it updates the abm config (location) for the main backend

(same applies for block cache on the ee repo)

domsj:
fixed by #789

Issue #775, opened 2017-08-07
Exception in alba_id_t: (Failure "attempting to use a resource from a pool which is being finalized")


domsj:
```
Aug 07 11:52:56 ctrl-221-3 alba[16810]: 2017-08-07 11:52:50 830713 +0200 - ctrl-221-3 - 16810/0000 - alba/maintenance - 8139446807 - info - Exception in alba_id_t: (Failure "attempting to use a resource from a pool which is being finalized"); backtrace:; Raised at file "format.ml", line 241, characters 41-52; Called from file "format.ml", line 482, characters 6-24
Aug 07 11:52:56 ctrl-221-3 alba[16810]: 2017-08-07 11:52:50 830729 +0200 - ctrl-221-3 - 16810/0000 - alba/maintenance - 8139446808 - info - Exception in alba_id_t: (Failure "attempting to use a resource from a pool which is being finalized"); backtrace:; Raised at file "format.ml", line 241, characters 41-52; Called from file "format.ml", line 482, characters 6-24
```

@pploegaert saw this in a maintenance process on his test env
and IIRC I also saw this on a proxy on the nuvolat setup

still to be further investigated

pploegaert:
fwiw (old alba version)

```
root@ctrl-221-3:/var/tmp# alba version
ee-1.4.2
git_revision: "tags/1.4.2-0-g093692a"
git_repo: "Not available"
compile_time: "03/05/2017 13:21:51 UTC"
machine: "172396896fd1 4.4.0-36-generic x86_64 x86_64 x86_64 GNU/Linux"
model_name: "Intel(R) Xeon(R) CPU E31220 @ 3.10GHz"
compiler_version: "4.03.0"
is_alba_test: false
dependencies:
arakoon_client                    1.9.17        Arakoon client
bz2                                0.6.0        [n/a]
cmdliner                          v1.0.0        Declarative definition of command line interfaces
core                       [unspecified]        
ctypes                            0.11.3        Combinators for binding to C libraries without writing any C.
ctypes.foreign                    0.11.3        Dynamic linking of C functions
kinetic-client                     0.0.6        Kinetic client
lwt                                3.0.0        Lightweight thread library for OCaml (core library)
lwt.unix                           3.0.0        Unix support for Lwt
oUnit                              2.0.0        Unit testing framework
ocplib-endian                        1.0        Optimised functions to read and write int16/32/64 from strings and bigarrays
ppx_deriving.enum                    4.1        [@@deriving enum]
[...]

PR #789, opened 2017-08-16
refresh cache backend abm config from proxy

domsj:


domsj:
fixes #777 and probably also fixes #775

PR #791, opened 2017-08-17
fix centos Dockerfile

domsj:


docker/centos-7/Dockerfile
```diff
@@ -1,22 +1,22 @@
...
-
 RUN useradd jenkins -u 1500 -g root
 
+RUN rpm --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
 RUN rpm -iUvh http://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-10.noarch.rpm
-RUN yum -y update && \
-    yum -y install python-pip openssl-devel gmp-devel which zlib-devel \
-                   ncurses-devel bzip2 cmake3 sudo
+
+RUN yum -y update \
```
domsj: FYI messed around with this part because I wanted to add the 'file' dependency (some configure scripts complain about it - but carry on nonetheless), and because the yum update & install should be done in the same docker step

Issue #740, opened 2017-06-16
proxy has too many fd open (deleted files)


dejonghb:
Found with version 1.3.16 on 2 nodes:

Example:
```
root@ftcmp03:~# ls -l /proc/9827/fd
total 0
lr-x------ 1 root root 64 Jun 13 18:27 0 -> /dev/null
lrwx------ 1 root root 64 Jun 13 18:27 1 -> socket:[92684397]
lrwx------ 1 root root 64 Jun 13 18:27 10 -> socket:[117865667]
lrwx------ 1 root root 64 Jun 13 18:27 100 -> socket:[112063349]
l-wx------ 1 root root 64 Jun 13 18:27 101 -> /tmp/68d1fa47-1d9d-4510-b24a-3d7fe532ded1-volume_configuration44R6VNI73cqc (deleted)
lrwx------ 1 root root 64 Jun 13 18:27 102 -> socket:[112057180]
l-wx------ 1 root root 64 Jun 13 18:27 103 -> /tmp/68d1fa47-1d9d-4510-b24a-3d7fe532ded1-volume_configuration44R6VNJllZ4D (deleted)
lrwx------ 1 root root 64 Jun 13 18:27 104 -> socket:[112062186]
l-wx------ 1 root root 64 Jun 13 18:27 105 -> /tmp/68d1fa47-1d9d-4510-b24a-3d7fe532ded1-volume_configuration44R6VN4YwpLo (deleted)
lrwx------ 1 root root 64 Jun 13 18:27 106 -> socket:[112083376]
l-wx------ 1 root root 64 Jun 13 18:27 107 -> /tmp/68d1fa47-1d9d-4510-b24a-3d7fe532ded1-volume_configuration5U4S9ugCwgzx (deleted)
lrwx------ 1 root root 64 Jun 13 18:27 108 -> socket:[112076784]
l-wx------ 1 root root 64 Jun 13 18:27 109 -> /tmp/68d1fa47-1d9d-4510-b24a-3d7fe532ded1-volume_configuration5U4S9uzCtzNf (deleted)
lrwx------ 1 root root 64 Jun 13 18:27 11 -> socket:[104390890]
lrwx------ 1 root root 64 Jun 13 18:27 110 -> socket:[112090353]
l-wx------ 1 root root 64 Jun 13 18:27 111 -> /tmp/68d1fa47-1d9d-4510-b24a-3d7fe532ded1-volume_configuration5U4S9ui6KSsG (deleted)
[...]

PR #793, opened 2017-08-18
bugfix for test automatic repair

domsj:
travis & jenkins should be all green again now :-)


domsj:
This may also fix/improve #461 and #740?

PR #799, opened 2017-08-23
remove broken objects from cache namespaces

### LINKED ISSUE #641: broken objects in cache namespaces should be cleaned up
It makes no sense to keep around metadata for objects for which no fragments are still available. Maintenance should clean these up.
Not sure if it makes sense to keep around objects for which some fragments (less than k) are still available?
(It could still serve some read requests, so they have at least some value...)

domsj:
fixes #641

### DIFF
--- a/ocaml/src/maintenance.ml
+++ b/ocaml/src/maintenance.ml
@@ -179,10 +179,10 @@ class client ?(retry_timeout = 60.)
          Lwt.return_unit)
         retry_timeout
 
-    method _is_cache_namespace ~namespace_id =
+    method private _is_cache_namespace ~namespace_id =
       nsm_host_access # get_namespace_info ~namespace_id >>= fun (namespace, namespace_info, _, _) ->
       let open Maintenance_config in
-      Hashtbl.exists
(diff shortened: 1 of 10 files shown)


ocaml/src/maintenance.ml
```diff
@@ -179,6 +179,14 @@ class client ?(retry_timeout = 60.)
          Lwt.return_unit)
         retry_timeout
 
+    method _is_cache_namespace ~namespace_id =
```
toolslive: private?

ocaml/src/maintenance.ml
```diff
@@ -179,6 +179,14 @@ class client ?(retry_timeout = 60.)
...
+      nsm_host_access # get_namespace_info ~namespace_id >>= fun (namespace, namespace_info, _, _) ->
+      let open Maintenance_config in
+      Hashtbl.exists
```
toolslive: This confused me (is this a function in Hashtbl?) until I saw you added it in the overload. 
It's ok, but may we want another name to indicate that it scans the hashtbl.

PR #805, opened 2017-08-24
ocaml.4.04.2 ... arakoon.1.9.22

toolslive:


docker/centos-7/Dockerfile
```diff
@@ -151,7 +152,7 @@ RUN cp /usr/local/lib/libgtest.a /lib64/
 
 # install specific arakoon.
 RUN git clone https://github.com/openvstorage/arakoon.git
-RUN cd arakoon && git pull && git checkout 1.9.21
+RUN cd arakoon && git pull && git checkout d4f257272b990c36949bc305485c30a672180980
```
domsj: it's cleaner to checkout tags/1.9.22 instead? (here and in the other dockerfiles)

docker/centos-7/Dockerfile
```diff
@@ -90,11 +90,12 @@ RUN eval `${opam_env}` && \
         depext \
         kinetic-client \
         tiny_json \
-        ppx_deriving \
+        ppx_deriving.4.1 \
         ppx_deriving_yojson \
+        base.v0.9.3 \
         core.v0.9.0 \
```
domsj: this has core.v0.9.0 while other dockerfiles have 0.9.1?

PR #808, opened 2017-08-25
More instrumentation

toolslive:


toolslive:
ok. backport before merge.

PR #817, opened 2017-09-18
Dirty reads

toolslive:


ocaml/src/alba_client_download.ml
```diff
@@ -24,7 +24,26 @@ let get_object_manifests'
...
+    let open Consistency in
+    match consistent_read with
+    | true  ->
+         client # get_object_manifests_by_name ~consistency:Consistent object_names
+    | false ->
+       begin
+         client # get_object_manifests_by_name ~consistency:No_guarantees object_names
+         >>= fun mfos ->
+         let has_nones =
+           let rec aux = function
```
domsj: Use `List.mem None mfos` here?

toolslive: right.

PR #818, opened 2017-09-19
local backend errors don't necessarily disqualify it

toolslive:


ocaml/src/osd_access.ml
```diff
@@ -357,11 +357,58 @@ class osd_access
...
+           Lwt_log.ign_info_f
+             "%s %S was unforeseen"
+             long_id
+             (Printexc.to_string exn)
+         in
+         match osd_info.kind with
+           | Asd     _   -> true, false, true
+           | Kinetic _   -> true, false, true
+           | Alba    _   -> true, false, false
+           | Alba2   _   -> true, false, false
```
domsj: this should be `true, false, true` for all kinds of osds?

toolslive: I think you want don't want to be that pessimistic for Alba & Alba2 osds

toolslive: ok.

PR #827, opened 2017-11-29
pin opam and packages

toolslive:


domsj:
the change regarding `opam_installer.sh` will be needed too? let's wait for travis to agree first ...

toolslive:
https://github.com/ocaml/opam/issues/3121

but travis also needs 14.04. So I'll do that too.

PR #829, opened 2018-01-31
Update maintenance documentation to remove etcd

wimpers:


domsj (approved):
for alba itself in fact nothing changed, which to me is a sign this info doesn't belong here...
but while it's here it's of course better to keep it correct ;-)

